PR #2, opened 2020-05-07
287 deploy to sandbox

avdata99:
For [#287](https://github.com/GSA/datagov-ckan-multi/issues/287)

 - Adds a circleCI file:
   + Run tests (which still not exists)
   + Deploy to sandbox on branch _master_ and if test are green
 - Add the jenkinsfile
 - Add the jenkis script
 - Added env to clarify jenkinsfile
```
environment {
    INVENTORY = 'inventories/sandbox',
    PLAYBOOK = 'catalog.yml'
  }
```

Requires:
 - validation for select the right playbook file and inventory
 - Check credentials


thejuliekramer:
@avdata99 this is now at the point where it is trying to call data.gov Jenkins and is throwing an error that we do not have access... @adborden would it be possible to get access here?

PR #10, opened 2020-05-12
Add DataGovTheme

avdata99:
For [#284](https://github.com/GSA/datagov-ckan-multi/issues/284)

image

A really simple first version of DataGovTheme running over this new environment
 - Extension installed in a new req file (just to dicuss a way to be more _deployable_
   + Just using reqs for each extension as they come (we need more work)
 - Using upstream versions of exts
 - Re-order bats test and add different files for test related to different extensions.
 - Removing patches
 - We still need to define requirements policy and probably discuss about Alpine (here [we have troubles](https://github.com/GSA/catalog.data.gov/pull/10/files#diff-3e8383a48cad28a1d5ff0858c686b780R15-R22) to install shapely python lib)

**It works**

Extras
It's not required here but I did a test for harvesting an easy data.json source:

image

image

Error in /datasets
image


ckan/Dockerfile
```diff
@@ -7,25 +7,6 @@ ARG TZ
...
-COPY patches ${APP_DIR}/patches
-
-RUN for d in $APP_DIR/patches/*; do \
-        if [ -d $d ]; then \
-            for f in `ls $d/*.patch | sort -g`; do \
-                cd $SRC_DIR/`basename "$d"` && echo "$0: Applying patch $f to $SRC_DIR/`basename $d`"; patch -p1 < "$f" ; \
-            done ; \
-        fi ; \
-    done
+RUN pip install -r requirements.txt
```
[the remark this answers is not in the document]
avdata99: We will change req file in a new issue

README.md
```diff
@@ -8,6 +8,10 @@ This is a local development harness for catalog.data.gov.
 
 Run `make build up` to build and run the docker images. 
 
+Add `127.0.0.1 ckan` to your `/etc/hosts` file
```
adborden, line 11: Is it possible to avoid this manual step?

ckan/Dockerfile
```diff
@@ -7,25 +9,24 @@ ARG TZ
...
+RUN cd /geos && \
+    ./configure && \
+    make -j 1 && \
+    make install
+
+# RUN echo "@comu http://dl-cdn.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories
+
+# RUN apk update && \
+#   apk add --update --force-overwrite "geos@comu" "geos-dev@comu" && \
+#   rm -rf /var/cache/apk/*
```
adborden, line 28: Are these comments necessary?

tests/test-base.bats
```diff
@@ -0,0 +1,34 @@
...
+load globals
+load test_helper
+
+
+@test "CKAN container is up" {
+  wait_for_app
+}
+
+@test "/user/login is up" {
+  test_view_login
```
adborden, line 12: Are these functions reusable? Why not just incline the code so we can see the test description with the test code?

tests/test_helper.bash
```diff
@@ -1,4 +1,254 @@
...
+          "contact_email": "tester@fake.com",
+          "contact_name": "Tester",
+          "modified": "2019-12-18",
+          "public_access_level": "public",
+          "publisher": "Department of the Interior",
+          "unique_id": "doi-'$RNDCODE'",
+          "title": "Test Dataset '$RNDCODE'",
+          "name": "test-dataset-'$RNDCODE'",
+          "program_code": "010:001"
+        }'
```
adborden, line 202: Maybe move this JSON payload to a test data fixture?

README.md
avdata99: No, deleted

ckan/Dockerfile
avdata99: No, deleted

tests/test-base.bats
avdata99: I transformed this function to be reusable [here](https://github.com/GSA/catalog.data.gov/pull/10/commits/4ef22440d2e556bc67ef2f477467e8c2dda8db30)

tests/test_helper.bash
avdata99: I moved JSON data to external files [here](https://github.com/GSA/catalog.data.gov/pull/10/commits/4ef22440d2e556bc67ef2f477467e8c2dda8db30)

README.md
avdata99: This change depends on your OS and requires root privileges.
The best approach here could be set _localhost:5000_ in all configs but I think we are going to have troubles using _datapusher_ and other network connections between containers.
I think it's okay to leave it as it is now.

PR #14, opened 2020-05-13
[cd] update deployment

adborden:
- Use consistent deploy-ci-app-<name>
- Limit Ansible plays to v2 (catalog-next) hosts


adborden:
After a few more tweaks, this is ready for review :)

PR #18, opened 2020-05-15
Harevst sources import/export

avdata99:
For [#302](https://github.com/GSA/datagov-ckan-multi/issues/302)

 - Read harvest sources from external CKAN instance
 - Use only active sources (`state=active`)
 - Iterate over this list, and create new harvest source in CKAN destination
 - Requires an organization to assign to all sources

Harvest imported: Note differences in UI.
At the left, the new UI (public and edit view)
At the right: current Catalog public view

image

Known errors:
 - Some sources come without config when we search so we miss it. E.g. Boise harvest source. 
    `https://catalog.data.gov/api/3/action/package_search?q=(type:harvest source_type:datajson)`
    This source use extras to save config

image


tools/harvest_source_import/remote_ckan/lib.py
```diff
@@ -0,0 +1,139 @@
...
+
+        logger.info('Creating source URL {}'.format(ckan_package['title']))
+
+        try:
+            req = requests.post(package_create_url, data=ckan_package, headers=headers)
+        except Exception as e:
+            error = 'ERROR creating harvest source: {} [{}]'.format(e, ckan_package)
+            return False, 0, error
+
+        content = req.content
```
starsinmypockets, line 118: Should this be in one of the `try` blocks in case there is no `req.content`

starsinmypockets: On line 120?

tools/harvest_source_import/requirements.txt
```diff
@@ -0,0 +1,2 @@
+# ckanapi>=4.3
+requests
```
starsinmypockets, line 2: Should this be versioned?

tools/harvest_source_import/README.md
```diff
@@ -0,0 +1,56 @@
+# Harvest source import
+
+
+```
+python import.py \
```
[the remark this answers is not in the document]
starsinmypockets: I mean in the README maybe separate the input (what the user will input) from output (what they should expect to see as a response)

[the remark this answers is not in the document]
avdata99: Fixed

tools/harvest_source_import/remote_ckan/lib.py
avdata99: Moved error detection before try to read _content_ [here](https://github.com/GSA/catalog.data.gov/commit/8f622b6c8d1ed0298a6893cc76dcd299756e1516)

tools/harvest_source_import/requirements.txt
avdata99: Fixed

tools/harvest_source_import/remote_ckan/lib.py
```diff
@@ -0,0 +1,139 @@
...
+            logger.info(f'  [{source_type}] Harvest source: {title} [{state}]')
+            if state == 'active':
+                yield hs
+
+        # if the page is not full, is the last one
+        if count + 1 < page_size:
+            return
+
+        # get next page   
+        yield from self.list_harvest_sources(source_type=source_type, start=start + page_size, page_size=page_size)
```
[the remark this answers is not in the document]
avdata99: We always get a page. Even if we don't have more data. In this cases we get an empty page and we don't get an error.

tools/harvest_source_import/remote_ckan/lib.py
```diff
@@ -0,0 +1,217 @@
...
+
+        ckan_package = {
+            'name': data['name'],
+            'owner_org': data['organization']['name'],
+            'title': data['title'],
+            'url': data['url'],
+            'notes': data['notes'],
+            'source_type': data['source_type'],
+            'frequency': data['frequency'],
+            'config': json.dumps(config, indent=4)
```
adborden, line 128: Is the JSON indent necessary? Will the extra whitespace be stripped when it's inserted into the database?

avdata99: No, it's just nice to see in the UI. I removed the indent.

avdata99:
> Looks good! Are you planning on adding unit tests?

I will add them in a new PR

PR #22, opened 2020-05-19
Use Pipenv to manage dependencies

woodt:
See README for details.


Makefile
```diff
@@ -34,10 +34,8 @@ test:
 	docker-compose -f docker-compose.yml -f docker-compose.test.yml build
 	docker-compose -f docker-compose.yml -f docker-compose.test.yml up --abort-on-container-exit test
 
-# TODO waiting for consensus package management solution
-# update-dependencies:
-#	docker-compose run --rm -T ckan pip --quiet freeze > requirements-freeze.txt
-#	docker-compose run --rm -T ckan pip install -r ${CKAN_HOME}/requirements-freeze.txt
+update-dependencies:
+	docker-compose run --rm -T ckan freeze-requirements.sh `id -u` `id -g` > ckan/requirements.txt
```
adborden, line 38:
```suggestion
	docker-compose run --rm -T ckan freeze-requirements.sh $(shell id -u) $(shell id -g) > ckan/requirements.txt
```

Please test this, my Make skills are :unamused:

README.md
```diff
@@ -32,10 +32,28 @@ The repository extends the Open Knowledge Foundation `ckan-dev:2.8` docker image
...
+
+To support sandbox installation via the ansible playbooks, there is a
+symbolic link `requirements-freeze.txt` that references
+`ckan/requirements.txt`.
+
+### Procedure for updating a dependency
+
+1.  Add/change the dependency in `requirements/Pipfile`
+2.  Run `make updates-dependencies`
+3.  Run `make build`
```
adborden, line 54: Could combine 2 & 3, maybe we want to include `test` to validate the change is good.

    make update-dependencies clean test

woodt:
So I like your idea about losing the requirements-freeze.txt symlink by updating the datagov-deploy playbook ... how about, in the interest of reduced coupling, I make that change (as a "chore") after we have everything deployed?  We can then (1) update the playbook to refer to `ckan/requirements.txt` and then (2) lose the symlink.  Those latter changes are not coupled, so they should be easy to move through.

woodt:
Oh ... when pipenv does have to generate a new lock file, it does take 2-3 minutes to run :-(.  But if the dependencies haven't changed and the lock file doesn't need to be updated, generating a `requirements.txt` file happens "instantly".

adborden:
> how about, in the interest of reduced coupling, I make that change (as a "chore") after we have everything deployed?

I'm all for this kind of thinking @woodt :+1:

PR #29, opened 2020-05-20
bats-support for testing sugar

adborden:
This adds a few BATS libraries to make testing a little less rough.

- Use the git version of BATS (our container is 2 years old)
- [bats-support](https://github.com/bats-core/bats-support) and [bats-assert](https://github.com/bats-core/bats-assert) provide `assert_*` methods
- Use the "pretty" formatter for BATS (`-p`)
- `db` wrapper around psql

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,8 +1,20 @@
-FROM bats/bats
+FROM bash as bats
 
-RUN apk add curl jq postgresql-client
+RUN apk add git && \
+  mkdir /bats && \
+  git clone https://github.com/bats-core/bats-core.git /bats && \
+  git clone https://github.com/bats-core/bats-support.git /bats/lib/bats-support && \
+  git clone https://github.com/bats-core/bats-assert.git /bats/lib/bats-assert
+
+FROM bash
+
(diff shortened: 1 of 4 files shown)


tests/test-base.bats
```diff
@@ -38,16 +38,42 @@ load test_helper
     --data 'login=not_a_user' \
     --data 'password=badpassword' 2>&1 >/dev/null)
 
-  echo "$output" >&2 # debug
+  log "$output"
   ! echo "$output" | grep -qi '^< set-cookie:.*auth_tkt'
 }
 
 @test "User can create org" {
```
starsinmypockets, line 45: There's no diff but we just moved these tests from test-helpers to test-base.

Dockerfile.test
```diff
@@ -1,8 +1,20 @@
-FROM bats/bats
+FROM bash as bats
```
starsinmypockets, line 1: Not sure why we don't use `bats/bats` image but sure

adborden: The bats/bats image is two years old at this point which bugs me. Once https://github.com/bats-core/bats-core/issues/266 is published to Docker Hub we can switch back.

PR #30, opened 2020-05-21
Adding tests

avdata99:
Done
 - Adding test for the import script
 - Added to run in CircleCI
 - Record request to pass the test without requiring real CKAN instances
   + GET from catalog.data.gov
   + POST to ckan:5000

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,19 +19,18 @@ jobs:
           command: bin/jenkins_build deploy-ci-app-catalog-next
   
   test_importer:
-    build:
-      working_directory: ~/tools/harvest_source_import 
-      docker:
-        - image: circleci/python:3.6.2-stretch-browsers
-      steps:
-        - checkout
-        - run:
-            name: tests 
-            command: |  
-              pip install -r dev-requirements.txt
-              flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-              flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-              python -m pytest --vcr-record=none tests/
+    working_directory: ~/tools/harvest_source_import 
+    docker:
+      - image: circleci/python:3.6.2-stretch-browsers
+    steps:
+      - checkout
+      - run:
+          name: tests 
+          command: |  
+            pip install -r dev-requirements.txt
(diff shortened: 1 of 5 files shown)


tools/harvest_source_import/README.md
```diff
@@ -69,3 +69,21 @@ remote_ckan.lib - Harvest source created OK Restore-the-gulf
...
+We use _pytest cassettes_ to save responses from orgin and destination CKAN instances.
+
+#### Record results
+
+We can run test against real CKAN instances to save each request reposnse (GET and POST)
+
+´´´
+python -m pytest --vcr-record=all tests/
+´´´
+#### Run test by used saved requests
```
starsinmypockets, line 84:
```suggestion
#### Run tests against saved requests
```

tools/harvest_source_import/tests/test_import_sources.py
```diff
@@ -0,0 +1,41 @@
+import pytest
```
starsinmypockets, line 1: I think this file should be called _import_test_sources.py_

avdata99: If I rename it pytest you won't see the file

.circleci/config.yml
```diff
@@ -17,6 +17,21 @@ jobs:
...
+      - checkout
+      - run:
+          name: tests 
+          command: | 
+            cd tools/harvest_source_import
+            pip install --upgrade pip 
+            pip install -r dev-requirements.txt
+            flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+            flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+            python -m pytest --vcr-record=none tests/
```
adborden, line 34: These steps are worth adding to a local subdir makefile so it's easy to test locally as well.

Also, with the addition of Pipenv, we should use a Pipfile instead of a requirements.txt.

avdata99: Done [here](https://github.com/GSA/catalog.data.gov/commit/2450e45e54669b29cdeb29d4c5c1291983b1195c)

adborden: Pipenv?

PR #31, opened 2020-05-21
Add update harvest source function and clean requests to CKAN

avdata99:
For [#302](https://github.com/GSA/datagov-ckan-multi/issues/302)

 - Update harvest sources when they exist
 - Clear repeated functions 
 - Tested locally
   + Add 5 sources
   + Then add 15 and get
```
Finished: 15 harvest sources. 10 Added, 5 already exists, 0 failed
````

### CODE REVIEW

File: tools/harvest_source_import/remote_ckan/lib.py [hunk 1]
```diff
@@ -194,34 +174,77 @@ def create_organization(self, data):
...
+        return True, req.status_code, None
+    
+    def get_config(self, data):
+        """ get config and extras from full data package and return a final str config """
+        config = data.get('config', {})
+        if type(config) == str:
+            config = json.loads(config)
+
+        # we can also have config in extras
+        # we don't get extras
```
adborden, line 229: Can you clarify the comment above? Maybe "We may not have extras in every response"?

avdata99: Fixed

avdata99:
> Any test updates?

No. I will update tests and then ask for review again

PR #45, opened 2020-05-28
Add missing plugin and move to postgis

### LINKED ISSUE #43: BUG, fail to show dataset from WAF source
Error
```
Viewing datasets of type "dataset" is not supported 
('Template spatial/snippets/dataset_map_sidebar.html cannot be found'). 
```
**How to reproduce**

Add a WAF source from the URL http://meta.geo.census.gov/data/existing/decennial/GEO/CPMB/boundary/2016gz/kml_aiannh_500/
Add config: 
```
{"validator_profiles": ["iso19139ngdc"], "private_datasets": false}
```
(you can also import WAF sources from the import script)
**This source was harvested last time in 2016: validate we expected to work**

Run the harvest process, you will get 2 datasets. You'll see the error trying to open any of these datasets.

image

#related to [#307](https://github.com/GSA/datagov-ckan-multi/issues/307)

-----

Same error for _New Mexico Resource Geographic Information System (NM RGIS)_ source
URL: http://rgismetadata.unm.edu/19115/collections/
Config
```
{"validator_profiles": ["iso19139ngdc"], "private_datasets": false}
```
We can get the datasets list but we can get inside

image

avdata99:
Fixes #43 

 - Move from PostgreSQL to Postgis container
 - Add missing Spatial plugin

Now we can see spatial datasets

image

### DIFF
--- a/.env
+++ b/.env
@@ -40,7 +40,7 @@ CKAN_SMTP_PASSWORD=pass
 CKAN_SMTP_MAIL_FROM=ckan@localhost
 
 # Extensions
-CKAN__PLUGINS=envvars image_view text_view recline_view datastore datapusher harvest ckan_harvester geodatagov datajson datajson_harvest geodatagov_miscs z3950_harvester arcgis_harvester geodatagov_geoportal_harvester waf_harvester_collection geodatagov_csw_harvester geodatagov_doc_harvester geodatagov_waf_harvester spatial_query datagovtheme
+CKAN__PLUGINS=envvars image_view text_view recline_view datastore datapusher harvest ckan_harvester geodatagov datajson datajson_harvest geodatagov_miscs z3950_harvester arcgis_harvester geodatagov_geoportal_harvester waf_harvester_collection geodatagov_csw_harvester geodatagov_doc_harvester geodatagov_waf_harvester spatial_metadata spatial_query datagovtheme
 
 # Harvest settings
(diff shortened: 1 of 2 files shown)


.env
```diff
@@ -51,4 +51,6 @@ CKAN__HARVEST__LOG_LEVEL=info
 CKAN__HARVEST__LOG_SCOPE=0
 
 CKANEXT__GEODATAGOV__BUREAU_CSV__URL=https://project-open-data.cio.gov/data/omb_bureau_codes.csv
-CKANEXT__GEODATAGOV__BUREAU_CSV__URL_DEFAULT=https://project-open-data.cio.gov/data/omb_bureau_codes.csv
\ No newline at end of file
+CKANEXT__GEODATAGOV__BUREAU_CSV__URL_DEFAULT=https://project-open-data.cio.gov/data/omb_bureau_codes.csv
+
+CKAN__SPATIAL__SRID=4326
```
woodt, line 56: Just out of curiosity, what is SRID?

woodt:
This looks fine.  Changes will also be needed to the ansible playbooks (new plugins + config file change.) See the README.md for details.

.env
avdata99: Is about which _map prjection_ you use to save spatial data. [More info](https://en.wikipedia.org/wiki/Spatial_reference_system)

PR #49, opened 2020-05-29
[WIP] Add archiver

starsinmypockets:
This appears to successfully initialize harvester but still throws helper error


ckan/docker-entrypoint.d/10-setup-archiver.sh
```diff
@@ -0,0 +1,6 @@
+echo "Setup Report database table"
+paster --plugin=ckanext-report report initdb --config=$CKAN_INI
+
+echo "Setup Archiver database table"
+paster --plugin=ckanext-archiver archiver init --config=$CKAN_INI
+paster --plugin=ckanext-report report initdb --config=$CKAN_INI
```
adborden, line 6: Is this line identical to line 2? Is it required?

PR #74, opened 2020-06-23
Replace "harvest" with "datagov_harvest"

avdata99:
Start using the `datagov_harvest` to override harvest source form creation
Now we are not able to create `waf-collection` harvester sources.

This is the [actual behaviour of catalog classic](https://github.com/GSA/datagov-deploy/blob/1cdb5dadecbc27a683d59e753dfeaab55bccc3ce/ansible/roles/software/ckan/catalog/ckan-app/defaults/main.yml#L16).

The form add this new fields:

image


.env
```diff
@@ -40,7 +40,7 @@ CKAN_SMTP_PASSWORD=pass
 CKAN_SMTP_MAIL_FROM=ckan@localhost
 
 # Extensions
-CKAN__PLUGINS=envvars image_view text_view recline_view datastore datapusher harvest ckan_harvester geodatagov datajson datajson_harvest geodatagov_miscs z3950_harvester arcgis_harvester geodatagov_geoportal_harvester waf_harvester_collection geodatagov_csw_harvester geodatagov_doc_harvester geodatagov_waf_harvester spatial_metadata spatial_query report archiver datagovtheme datagovcatalog
+CKAN__PLUGINS=envvars image_view text_view recline_view datastore datapusher datagov_harvest ckan_harvester geodatagov datajson datajson_harvest geodatagov_miscs z3950_harvester arcgis_harvester geodatagov_geoportal_harvester waf_harvester_collection geodatagov_csw_harvester geodatagov_doc_harvester geodatagov_waf_harvester spatial_metadata spatial_query report archiver datagovtheme datagovcatalog
```
thejuliekramer, line 43: do we want `harvest` AND `datagov_harvest` or just `datagov_harvest` here

avdata99: We have a _duplicate dataset type_ error if we use both. Catalog classic just lose this one

Issue #42, opened 2020-05-27
Bug: Harvest source importer misses some metadata


avdata99:
## Bug report
Expected Behavior

Actual Behavior

## Details/Tasks
- Related to [#302](https://github.com/GSA/datagov-ckan-multi/issues/302) and [#312](https://github.com/GSA/datagov-ckan-multi/issues/312)

## Acceptance Criteria
After we import harvest sources we need to validate we have all the information we expect
 - [x] Check if we have the same config as the original source (sometimes the config come as _extras_)
 - [x] Check if we have the same extras as the original source (if we detect _config_ inside _extras_ we move it to config and remove from extras)
 - [x] Check the organization is the same and has the same image included.
 - [x] in the old/classic catalog app (with CKAN 2.3) when we create a harvest soure we need to pick federal or not-federal source. This info seems not been captured while we import. Is this private data?: **No, this info is saved as validator_schema in extra['config']. If validator_schema is empty the source is federal** 

At the [final report](https://docs.google.com/spreadsheets/d/1LgVvgiUC5Vc-OQaSmtvQNlTFLhs3Mho65ekggYEiYh4/edit?ts=5e4177f8#gid=1667396946) about harvest sources in the new catalog we found some error that make us think that we fail while import this harfest sources.

Some sources faileds with validation errors:

| source     | url | config | added | updated | nulls | errors | 
|------------- | --- | -------- | --------- | ---------- | ------- | ------- | 
epa-sciencehub | https://pasteur.epa.gov/metadata.json | {"private_datasets": "False"} | 1665 | 0 | -1 | Title: EPA True NO2 ground site measurements. 1 Error(s) Found. ### ERROR 1: 'distribution':[{'accessURL': 'http://www-air.larc.nasa.gov/cgi-bin/ArcView/disco ...
mcc-data-json | https://data.mcc.gov/data.json | {"private_datasets": "False"} | 199 | 0 | -1 | Title: Ghana - Agriculture - Land Tenure; 1 Error(s) Found. ### ERROR 1: 'distribution':[{'downloadURL': 'http://microdata.worldbank.org/index.php/cata ...
[...]

PR #78, opened 2020-07-02
WIP: Expand imported sources info

avdata99:
Related to #42 

Improvements:
 - Get Organization extras (is required for `email_list` to send notifications)
    + Test added for organization's extras
 - Save harvest sources and organization data as JSON to analyze locally.
 - Manually tested: we now get full info about Organizations

image

### DIFF
--- a/tools/harvest_source_import/remote_ckan/lib.py
+++ b/tools/harvest_source_import/remote_ckan/lib.py
@@ -12,6 +12,7 @@ class RemoteCKAN:
         self.user_agent = user_agent
         self.errors = []
         self.harvest_sources = {} 
+        self.organizations = {}
         logger.debug(f'New remote CKAN {url}')
         # save results temp data into a new folder
         self.temp_data = temp_data
@@ -137,11 +138,13 @@ class RemoteCKAN:
             self.errors.append(error)
             # yield incomplete version
             self.save_temp_json('organization', org.get('name', org.get('id', 'unnamed')), org)
+            self.organizations[org['name']] = org
             return org
         else:
             response = response.json()
             org = response['result']
             self.save_temp_json('organization', org.get('name', org.get('id', 'unnamed')), org)
+            self.organizations[org['name']] = org
             return org
 
     def get_request_headers(self, include_api_key=True):
--- a/tools/harvest_source_import/remote_ckan/lib.py
+++ b/tools/harvest_source_import/remote_ckan/lib.py
@@ -1,4 +1,5 @@
 import json
+import os
 import requests
 from remote_ckan.logs import get_logger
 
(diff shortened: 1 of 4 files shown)


tools/harvest_source_import/remote_ckan/lib.py
```diff
@@ -75,7 +81,7 @@ def list_harvest_sources(self, source_type=None, start=0, page_size=100, limit=0
             logger.info(f'  [{source_type}] Harvest source: {title} [{state}]')
             if state == 'active':
                 # We don't get full harvest soure info here. We need a custom call
-                yield self.get_full_harvest_source(hs)
+                yield self.get_full_harvest_source(hs)
```
woodt, line 84: some trailing whitespace here that you probably don't need :-)

tools/harvest_source_import/remote_ckan/lib.py
```diff
@@ -101,11 +107,45 @@ def get_full_harvest_source(self, hs):
...
+            logger.error(error)
+            self.errors.append(error)
+            # yield incomplete version
+            self.save_temp_json('organization', org.get('name', org.get('id', 'unnamed')), org)
+            self.organizations[org['name']] = org
+            return org
+        else:
+            response = response.json()
+            org = response['result']
+            self.save_temp_json('organization', org.get('name', org.get('id', 'unnamed')), org)
```
woodt, line 146: Line 146-148 are dups of 140-142 ... this could be refactored slightly to avoid the duplication.

tools/harvest_source_import/tests/test_import_sources.py
```diff
@@ -11,12 +11,15 @@ def test_load_from_url():
 
     ckan = RemoteCKAN(url='https://catalog.data.gov')
     ckan.set_destination(ckan_url='http://ckan:5000',
-                         ckan_api_key='7564fcf7-1e79-4e03-a052-f94979051770')
+                         ckan_api_key='783b1397-3a9f-4ffe-81c3-c67cc85ef5f1')
```
woodt, line 14: Curious: why did the API key change?

tools/harvest_source_import/remote_ckan/lib.py
avdata99: Yes, you are right. Fixed [here](https://github.com/GSA/catalog.data.gov/pull/78/commits/42ef6d80f0604512ae273a1419e32c03b8742d66)

tools/harvest_source_import/remote_ckan/lib.py
avdata99: Yes, you are right. Fixed [here](https://github.com/GSA/catalog.data.gov/pull/78/commits/42ef6d80f0604512ae273a1419e32c03b8742d66)

tools/harvest_source_import/remote_ckan/lib.py
avdata99: fixed

tools/harvest_source_import/tests/test_import_sources.py
avdata99: A way to avoid require a full CKAN running is to save the request responses with Pytest VCR cassettes.
Each API key is the valid API keys while I test this (changes every time I clean my environment). I need to update to fit the exact request. [Here](https://github.com/GSA/catalog.data.gov/blob/42_expand_imported_sources_info/tools/harvest_source_import/tests/cassettes/test_load_from_url.yaml#L989:L989) you can see where this API KEY is used.

woodt: That's what I thought ... but shouldn't the changed pytest cassette be part of this pull request, then?

PR #80, opened 2020-07-15
Allow list while import

avdata99:
Related to #42 
This PR:
 - Allow import sources from a list of names (like `check_harvest_sources` script)
 - Add a `update_organization` function

### DIFF
--- a/tools/harvest_source_import/import_harvest_sources.py
+++ b/tools/harvest_source_import/import_harvest_sources.py
@@ -15,6 +15,7 @@ parser.add_argument("--user_agent", type=str, default='CKAN-harvest-source-impor
 parser.add_argument("--destination_url", type=str, default='http://ckan:5000', help="CKAN destination instance URL")
 parser.add_argument("--destination_api_key", type=str, help="CKAN destination instance API KEY")
 parser.add_argument("--limit", type=int, default=0, help="Limit the amount of Harvest sources to import")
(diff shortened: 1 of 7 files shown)


tools/harvest_source_import/import_harvest_sources.py
```diff
@@ -1,24 +1,56 @@
...
 ckan = RemoteCKAN(url=args.origin_url, user_agent=args.user_agent)
 ckan.set_destination(ckan_url=args.destination_url, ckan_api_key=args.destination_api_key)
 
-for hs in ckan.list_harvest_sources(source_type=args.source_type, limit=args.limit):
+# define the final list of sources to import (from type o a list)
+sources_to_import = []
+
+if args.names is not None:
+    # we get a list of names from a file or list of source names 
+    if os.path.isfile(args.names):
```
starsinmypockets, line 30: should this get wrapped in try/catch to catch filesystem failure, or is it better to throw uncaught exception here?

tools/harvest_source_import/import_harvest_sources.py
```diff
@@ -1,24 +1,56 @@
...
+        f = open(args.names)
```
starsinmypockets, line 31: actually here ^^

avdata99: This file is opened only if it exists. So, in ths especific case we don't need to catch an error.

PR #84, opened 2020-07-22
Remove patch

avdata99:
- Remove patch
 - Use new fork
 - Update dependencies


woodt:
It looks like there is  a problem with the CI build, though.

Issue #95, opened 2020-08-03
Error importing WAF harvest sources


avdata99:
Source with config: `'{"validator_profiles": ["fgdc_minimal"]}` fails with `Unknown validation profile(s): fgdc_minimal`

Harvest source failing: [fema-r10](https://catalog.data.gov/harvest/fema-r10)

## How to reproduce

Run the importer for this source:
```
python import_harvest_sources.py --names=fema-r10 --destination_api_key=b396d3d9-1f24-4b72-92c1-78f981009a92
```

Full importer error

```
****** creating fema-r10: 7 of 289 sources
remote_ckan.lib - Get organization data https://catalog.data.gov/api/3/action/organization_show {'id': 'fema-gov'}
remote_ckan.lib - organization fema-gov saved at /checks
remote_ckan.lib - New extra found at org email_list=ITHelp@riskmapcds.com
david.sakoda@associates.fema.dhs.gov
crystal.carter@gsa.gov

remote_ckan.lib - New extra found at org organization_type=Federal Government
remote_ckan.lib - Creating organization fema-gov
remote_ckan.lib - Get organization data https://catalog-next.sandbox.datagov.us/api/3/action/organization_show {'id': 'fema-gov'}
remote_ckan.lib - organization fema-gov saved at /checks
remote_ckan.lib - Request OK https://catalog-next.sandbox.datagov.us/api/3/action/organization_update
remote_ckan.lib - Config found in extras: {'value': '{"validator_profiles": ["fgdc_minimal"], "private_datasets": false}', 'key': 'config'}
remote_ckan.lib - Creating harvest source FEMA-R10
remote_ckan.lib - ERROR status: 409
	 content:b'{"help": "https://catalog-next.sandbox.datagov.us/api/3/action/help_show?name=harvest_source_create", "success": false, "error": {"__type": "Validation Error", "config": ["Error parsing the configuration options: Unknown validation profile(s): fgdc_minimal"]}}'
	 sent: {'name': 'fema-r10', 'owner_org': 'fema-gov', 'title': 'FEMA-R10', 'url': 'https://hazards.fema.gov/filedownload/metadata/R10/', 'notes': '', 'source_type': 'waf', 'frequency': 'WEEKLY', 'config': '{"validator_profiles": ["fgdc_minimal"], "private_datasets": false}', 'extras': [], 'id': 'fd47c02e-3ef2-42fb-b4e5-40560b136a08'}
```

avdata99:
[...]

PR #96, opened 2020-08-03
Fix WAF sources import

avdata99:
Related to #95 

 - Remove validator_proflies from config, avoid Spatial to fail with regular profiles (we are using custom profiles)


tools/harvest_source_import/remote_ckan/lib.py
```diff
@@ -498,9 +498,6 @@ def get_package_from_data(self, data):
         if cfg.get('private_datasets', None) is not None:
             ret['private_datasets'] = cfg['private_datasets']
             
-        if cfg.get('default_groups', None) is not None:
```
thejuliekramer: @adborden I remove this because it was causing an error in the test where we actually use default groups - setting these here might break other harvest types depending on which they use - we might want to consider only adding what we need as we go

PR #127, opened 2020-09-25
Updates for harvest source report

adborden:
https://github.com/GSA/datagov-ckan-multi/issues/454

- Fix an issue with the source_type handling
- Add errors reported to the harvest report


adborden:
I've asked @avdata99 to look over the test case and see if we can use cassettes for this instead.

avdata99 (approved):

tools/harvest_source_import/tests/test_list_sources.py
```diff
@@ -0,0 +1,92 @@
...
+        results[hs['name']] = hs
+
+    assert total == 2
+    assert results['doi-open-data']['url'] == 'https://data.doi.gov'
+    assert results['doi-open-data']['status']['job_count'] == 1
+
+
+def test_requests_sent_for_ckan_sources():
+    cass = vcr.cassette.Cassette.load(path='tests/cassettes/test_list_ckan_sources.yaml')
+    search_request = cass.requests[0]
```
adborden, line 33: How do you know that the request's we're inspecting is the actual request and not the request listed in the [test data](https://github.com/GSA/catalog.data.gov/pull/127/files#diff-aafb6264c1e5675c7ea4ccaab2c0bf43R123)?

I would expect to see a `RemoteCKAN.list_harvest_sources` call within the test. Otherwise it looks like we're not running any code under test.

avdata99: I assume the list of requests in the `test_list_ckan_sources.yaml` file are ordered. The first one is the `package_search` for the first page of results.

The `list_harvest_sources` function [is called](https://github.com/GSA/catalog.data.gov/blob/bugfix/harvest-source-type/tools/harvest_source_import/tests/test_list_sources.py#L20) in the tests. This call saved the yaml file (with the same name as the test function).

You can delete the cassettes and run the test again and it will continue working.

avdata99: I will merge, we can add new changes in a new PR if needed.

PR #131, opened 2020-10-02
Hot fix make commands

avdata99:
Just fix `make` commands


danmayol (changes requested):
@avdata99 Fixing the typo I see :-) 
Looks like the last make command could use the same fix (harvets-check-finished-jobs), otherwise looks good!

avdata99:
I added a second commit @danmayol, ready to re-review

danmayol (approved):
Looks great!

PR #133, opened 2020-10-05
Run integrations test for our extensions

avdata99:
Each extension has its own test.
In the current environment if a change in _extension A_ breaks something in _extension B_ we didn't capture.
We need an environment that test changes in one extension that affect others.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,7 @@ jobs:
       - run: make build ci
       - run: make test
       - run: make lint-all
+      - run: make test-extensions
 
   deploy-sandbox:
     docker:
--- a/Makefile
+++ b/Makefile
@@ -98,3 +98,24 @@ harvest-gather-queue:
 
 harvest-check-finished-jobs:
 	docker-compose exec ckan paster --plugin=ckanext-harvest harvester run
+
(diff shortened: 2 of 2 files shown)


adborden (approved):
🤷  this doesn't feel quite right to me. I feel like if we want to test behavior of two extensions together, we should write an explicit test case for that. Running the unit tests is sort of a catch all because CKAN doesn't have good tools available to control the test environment. @avdata99 can you link to specific issues that this would've caught?

If we're running unit tests for these extensions, why not other extensions? Why not run unit tests for dependencies like `requests`? Why not run unit tests for postgresql?

Anyway, let's consider this an experiment and try it out. We'll see what kinds of issues this helps us identify. But let's not use this as an excuse to not write integration tests.

avdata99:
@adborden The GeoDataGov extension affects other extensions. Sometimes I run manually datajson test after a change it. We also created months ago datajson tests inside GoeDataGov.
The CI environment for each extension is different. 
This is only a way to use the tools that we already have in an integrated way.

I do not pretend to replace integration test, I just think that if some change en GeoDataGov breaks we can capture before deploy.
I agree that we will need more integration tests

PR #136, opened 2020-10-07
Increase verbosity and fix poetry version error

avdata99:
Related to [Multi#477](https://github.com/GSA/datagov-ckan-multi/issues/477)

### DIFF
--- a/ckan/Dockerfile
+++ b/ckan/Dockerfile
@@ -34,7 +34,7 @@ RUN cd ${SRC_DIR}/ckan && \
 RUN sed -i '/GSA\/ckan\.git/d' ${APP_DIR}/requirements.txt
 
 RUN pip install -r requirements.txt
-RUN pip install poetry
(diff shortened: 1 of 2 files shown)


woodt (approved):
This looks fine to me.  It will be *very* verbose, though.  Maybe just one or two "-v" if you really want more output?

danmayol (commented):
Change to pin the version of poetry looks good to me as well. 
Unsure about the verbosity change though. In some quick tests (so perhaps I missed something) increasing the verbosity did not seem to catch this error for me. The output made it as far as "derived: ckan-catalog" and then hung there. So the verbosity might only increase for successful tests (verified increased output with the pinned version) and not necessarily be useful for failures of this nature. So to Tom's point, maybe add a lesser verbosity level. a single 'v' did not seem to show the same dependency step, two did. So maybe use "-vv"?

PR #149, opened 2020-10-21
Feature/new relic

jbrown-xentity:
Related to https://github.com/GSA/datagov-deploy/issues/2263.
To clean up various aspects of files that are overwriting files from upstream.


ckan/setup/server_start.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
adborden, line 3: Doesn't need to be bullet proof. Should be easy to read.
```suggestion
DIR="$(dirname "${BASH_SOURCE[0]}")"
```

ckan/setup/server_start.sh
```diff
@@ -0,0 +1,10 @@
...
+
+if test -f "$DIR/.env"; then
+    set -o allexport; source $DIR/.env; set +o allexport
+fi
+
+# Run web application
+$CKAN_PY_ENV/newrelic-admin run-program $CKAN_PY_ENV/gunicorn --worker-class gevent --paste $CKAN_INI "$@"
```
adborden, line 10: Instead of CKAN_PY_ENV, how about we just assume the caller has already setup the path correctly and/or activated the virtualenv?

jbrown-xentity: I hit a blocker due to [not being able to setup the virtual environment via Ansible](https://stackoverflow.com/questions/20040141/ansible-command-from-inside-virtualenv). I could [setup the path variable in supervisor](https://stackoverflow.com/questions/15202760/how-to-use-virtualenvwrapper-in-supervisor) to contain the virtualenv, is that a reasonable solution?

jbrown-xentity: I went ahead and removed that variable/path and we can discuss changes/details in the [datagov-deploy PR](https://github.com/GSA/datagov-deploy/pull/2317)

.env
```diff
@@ -68,4 +68,13 @@ CKAN___CELERY_RESULT_BACKEND=redis
...
+CKAN___REDIS_CONNECT_RETRY=True
+
+# New Relic
+NEW_RELIC_LICENSE_KEY=
+NEW_RELIC_APP_NAME=catalog-next
+NEW_RELIC_MONITOR_MODE=false
+NEW_RELIC_LOG=/var/log/new_relic.log
+NEW_RELIC_LOG_LEVEL=info
+NEW_RELIC_HOST=gov-collector.newrelic.com
+CKAN_PY_ENV=/usr/bin
```
adborden, line 80: No longer used?

jbrown-xentity: Fixed

adborden (approved):

avdata99 (approved):

Issue #153, opened 2020-10-26
update-dependencies is not working


avdata99:
I updated the `datagovtheme` extension today and after run `make update-dependencies` I didn't see the new version in the requirement file

Logs:

```
make update-dependencies 
docker-compose run --rm -T ckan freeze-requirements.sh 1000 1000
Starting catalogdatagov_redis_1 ... done
Starting catalogdatagov_db_1    ... done
Running poetry lock ...
Creating virtualenv ckan-catalog-C0o-n6lY-py2.7 in /root/.cache/pypoetry/virtualenvs
Using virtualenv: /root/.cache/pypoetry/virtualenvs/ckan-catalog-C0o-n6lY-py2.7
Updating dependencies
Resolving dependencies...
   1: fact: ckan-catalog is 0.1.0
   1: derived: ckan-catalog
   1: fact: ckan-catalog depends on ckanext-datagovcatalog (0.0.1 git)
   1: fact: ckan-catalog depends on repoze.who (2.3)
   1: fact: ckan-catalog depends on psycopg2 (2.7.3.2)
   1: fact: ckan-catalog depends on Flask (0.12.4)
   1: fact: ckan-catalog depends on OWSLib (0.8.6)
   1: fact: ckan-catalog depends on cffi (1.12.3)
   1: fact: ckan-catalog depends on pyutilib.component.core (4.6.4)
   1: fact: ckan-catalog depends on fanstatic (0.12)
   1: fact: ckan-catalog depends on messytables (>=0.15.2)
   1: fact: ckan-catalog depends on gevent (*)
   1: fact: ckan-catalog depends on ckanext-report (0.1 git)
   1: fact: ckan-catalog depends on GeoAlchemy (>=0.6)
   1: fact: ckan-catalog depends on newrelic (*)
   1: fact: ckan-catalog depends on pyparsing (>=2.1.10)
   1: fact: ckan-catalog depends on pysolr (3.6.0)
   1: fact: ckan-catalog depends on xlrd (>=1.0.0)
   1: fact: ckan-catalog depends on ckan (2.8.6 git)
   1: fact: ckan-catalog depends on sqlparse (0.2.2)
   1: fact: ckan-catalog depends on lxml (>=2.3)
   1: fact: ckan-catalog depends on rq (0.6.0)
   1: fact: ckan-catalog depends on ckanext-googleanalyticsbasic (0.1 git)
   1: fact: ckan-catalog depends on passlib (1.7.3)
[...]

PR #154, opened 2020-10-26
Fix update-dependencies

avdata99:
Related to #153


woodt (approved):
Dependencies appear to be updated correctly.  As long as our tests pass, I am good with this change.

PR #202, opened 2021-01-12
Temporary optional fix to update SQLAlchemy

avdata99:
Related to [Multi#553](https://github.com/GSA/datagov-ckan-multi/issues/553)

Creating PR to test. Probably not a final solution


avdata99:
This PR is a test to use new SQLAlchemy version
Closing since this will be covered with #203

PR #216, opened 2021-01-29
Update saml2auth and DataGovTheme

avdata99:


FuhuXia:
Is there any reason we dont use branch names?

avdata99:
> Is there any reason we dont use branch names?

We are using them [here](https://github.com/GSA/catalog.data.gov/blob/master/requirements/pyproject.toml#L24). This PR is just an automatic frozen version of each library.

FuhuXia (approved):

PR #240, opened 2021-02-11
Feature/GitHub actions

jbrown-xentity:
Integrate github actions for automated deployment to cloud.gov.

Some circleCi configuration will live on (docker publishing, for instance), will address in https://github.com/GSA/datagov-deploy/issues/2791.


adborden (approved):

.github/workflows/commit.yml
```diff
@@ -0,0 +1,138 @@
...
+        run: docker save -o ckan/image-catalog.tar datagov/catalog.data.gov:latest
+      - name: Archive Solr Docker image
+        run: docker save -o ckan/image-solr.tar datagov/catalog.data.gov.solr:latest
+      - name: Archive DB Docker image
+        run: docker save -o ckan/image-db.tar datagov/catalog.data.gov.db:latest
+      - uses: actions/upload-artifact@v2
+        with:
+          name: docker-images
+          path: ckan/
+          if-no-files-found: error
```
adborden, line 27: Do we need to archive these? Where is the docker push happening?

adborden:
What's left in CircleCI? Can that be stubbed out?

.github/workflows/commit.yml
jbrown-xentity: It's not yet, but I was able to do that part quickly. Created an issue (see description of this PR) to backlog this work.

.github/workflows/commit.yml
```diff
@@ -0,0 +1,138 @@
...
+      - name: Archive Catalog Docker image
+        run: docker save -o ckan/image-catalog.tar datagov/catalog.data.gov:latest
+      - name: Archive Solr Docker image
+        run: docker save -o ckan/image-solr.tar datagov/catalog.data.gov.solr:latest
+      - name: Archive DB Docker image
+        run: docker save -o ckan/image-db.tar datagov/catalog.data.gov.db:latest
+      - uses: actions/upload-artifact@v2
+        with:
+          name: docker-images
+          path: ckan/
```
adborden, line 26: Instead of archiving this directory, which has some code in it in addition to the tar archives, let's just artifact the individual files.
```suggestion
          path: |
            ckan/image-catalog.tar
            ckan/image-solr.tar
            ckan/image-solr.tar
```

adborden:
@jbrown-xentity for the harvest source import tool tests, the error is that it can't find a version for pytest. This is because pytest >=5.x no longer supports python 2... and if you check the CircleCI config, you'll see it's building using a python 3 container. I'll submit a PR.

I also updated the protected branch rules to include GH actions.

PR #241, opened 2021-02-12
Fix harvest_source_import test on GH actions

adborden:
The harvest source import tool is only built for python 3 and we need to make sure to setup GH actions with python 3.

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -40,6 +40,11 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):

jbrown-xentity:
Not sure why the jobs from inventory are required on catalog; something weird happening here on required checks @adborden?

adborden:
Oops, my mistake. The jobs are showing up because I copy pasted them on another branch while debugging deploy and then I just assumed since they were GH checks that they should be required. Fixed now!

PR #242, opened 2021-02-12
Add db inits for extension and gunicorn server

jbrown-xentity:
Related to https://github.com/GSA/datagov-deploy/issues/2604

The deploy worked, however the smoke test failed because the server was 7 seconds from actually serving traffic. Adding the db initializations causes the startup script to take 17 seconds (was already waiting 10 seconds before running the smoke test)...

Having that much down time isn't great whenever we deploy. Does anyone have a better process for running those db initializations? Should we run them as a thread, similar to how we [run the seeding script for inventory](https://github.com/GSA/inventory-app/blob/main/start.sh#L65)

Blocked by https://github.com/GSA/catalog.data.gov/pull/243; need secret service integration before adding necessary New Relic variables for start script... See [secret parsing](https://github.com/GSA/inventory-app/blob/main/cfstart.sh#L39) and [environment defaults](https://github.com/GSA/inventory-app/blob/main/manifest.yml#L23-L27) on inventory.

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -58,7 +58,7 @@ jobs:
           if-no-files-found: error
 
   deploy-development:
-    if: github.ref == 'refs/heads/develop'
+    # if: github.ref == 'refs/heads/develop'
     name: deploy (development)
     environment: development
     runs-on: ubuntu-latest
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -58,7 +58,7 @@ jobs:
           if-no-files-found: error
 
   deploy-development:
-    # if: github.ref == 'refs/heads/develop'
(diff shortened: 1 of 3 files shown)


adborden:
> Related to [GSA/datagov-deploy#2604](https://github.com/GSA/datagov-deploy/issues/2604)
> Having that much down time isn't great whenever we deploy. Does anyone have a better process for running those db initializations? Should we run them as a thread, similar to how we [run the seeding script for inventory](https://github.com/GSA/inventory-app/blob/main/start.sh#L65)

I would expect this is a one-time thing to work around. Migration and initialization should be much faster on the second run (should be a noop). We should **not** run them in a background job or thread. If the migration fails, then we definitely want our deployment to fail so that we can rollback and correct it.

adborden (approved):

PR #269, opened 2021-03-24
Build and tag images with fcs

jbrown-xentity:
Will tag builds on the fcs branch with `fcs` in dockerhub; will allow the builds to remain separate. `main` will still tag with `latest`.
Related to https://github.com/GSA/datagov-deploy/issues/2791

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,13 +10,13 @@ jobs:
           command: make build
       - run:
           name: Archive Catalog Docker image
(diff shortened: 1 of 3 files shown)


adborden:
I would leave fcs as-is and only make this change on `main`. `fcs` can still push to hub.docker.com and won't effect builds on `main`.

jbrown-xentity:
Agreed. Closing.

PR #298, opened 2021-05-25
Use solr 6.6.x for testing

adborden:
https://github.com/GSA/datagov-deploy/issues/3245

Use the latest 6.x release which includes fixes for known security
vulnerabilities.


jbrown-xentity (approved):

nickumia-reisys:
The 'API Key' not being available was an issue that was caused by the solr process failing.. If we want the updated solrconfig.xml, we should merge main back into here and then pull the new solrconfig.xml.

jbrown-xentity:
Implemented in different PR, closing.
https://github.com/GSA/catalog.data.gov/blob/main/solr/Dockerfile#L1

PR #300, opened 2021-05-26
Remove unnecessary pg setup

jbrown-xentity:
Gives the following errors on a new uninitialized DB:
db_1           | initdb: directory "/var/lib/postgresql/data" exists but is not empty
db_1           | If you want to create a new database system, either remove or empty
db_1           | the directory "/var/lib/postgresql/data" or run initdb
db_1           | with an argument other than "/var/lib/postgresql/data".

The setup for system access is unnecessary now, removing it leaves the `/data` folder empty and the initialization happens normally.

### DIFF
--- a/postgresql/Dockerfile
+++ b/postgresql/Dockerfile
@@ -1,9 +1,6 @@
 FROM mdillon/postgis:10-alpine
 MAINTAINER Open Knowledge International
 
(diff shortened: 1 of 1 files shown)


adborden (approved):
👍  still scratching my head why it's not affecting everyone, including CI without this fix.

PR #301, opened 2021-05-26
Update circleci-build-catalog-next.bash

adborden:
This shared script is used for the legacy "travis-build script" style tests in other repos. Since these are legacy and are only testing CKAN 2.8 with our legacy catalog/inventory setup, we should be using the fcs branch instead of master (or main) which are targeting cloud.gov.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,21 +47,22 @@ jobs:
     steps:
       - checkout
       - run:
-          name: tests 
+          name: tests
           command: make test-import-tool
-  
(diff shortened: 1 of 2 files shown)


mogul (dismissed):

jbrown-xentity (approved):

adborden:
the start_ckan_28 test should be passing once https://github.com/GSA/catalog.data.gov/pull/304 is merged.

mogul (approved):

PR #304, opened 2021-05-26
Use tag for CKAN (fcs)

adborden:
We hit a bug that was introduced on the CKAN 2.8 branch but **not** on the 2.8.8
tag. Let's stick to the tags which are officially released.


adborden:
This is kind of funny how the test dependencies bleed into each other. I think merging this `main` PR https://github.com/GSA/catalog.data.gov/pull/303 will fix the test on FCS, since the script is pointing to master (which presumably is redirected to `main`)?

mogul (approved):

PR #315, opened 2021-06-09
Solr upgrade

nickumia-reisys:
The upgraded version of solr does fix: https://github.com/GSA/datagov-deploy/issues/3245

A demonstration can be shown, but with the older version of solr, external entity resolution works and with the new version, external entity resolution becomes unsupported.  

This may be an immature pull request depending on the two following reasons:
1) Updating the config.xml(s) from 6.0.0 to 6.6.6 was only done because it seems appropriate, I'm not sure if those were frozen at 6.0.0 for a reason.
2) I didn't make a test to show the vulnerability was fixed.  It is possible, but I didn't think about doing it before re-reading the Contributing best practices.  Semi-related, I'm also not sure about sensitivity of information, so I left it out until we talk about it.


solr/Dockerfile
```diff
@@ -15,15 +16,19 @@ RUN mkdir -p /opt/solr/server/solr/$SOLR_CORE/data
...
+https://raw.githubusercontent.com/apache/lucene-solr/releases/lucene-solr/6.6.6/solr/server/solr/configsets/basic_configs/conf/protwords.txt \
+https://raw.githubusercontent.com/apache/lucene-solr/releases/lucene-solr/6.6.6/solr/server/solr/configsets/data_driven_schema_configs/conf/elevate.xml \
 /opt/solr/server/solr/$SOLR_CORE/conf/
 
 # Create Core.properties
 RUN echo name=$SOLR_CORE > /opt/solr/server/solr/$SOLR_CORE/core.properties
 
 # Giving ownership to Solr
-RUN chown -R $SOLR_USER:$SOLR_USER /opt/solr/server/solr/$SOLR_CORE
+RUN mkdir /opt/solr/server/solr/ckan/data/index
```
FuhuXia:
```suggestion
RUN mkdir /opt/solr/server/solr/$SOLR_CORE/data/index
```

FuhuXia: Use the parameter instead of hard coded `ckan`

FuhuXia (approved):

jbrown-xentity (approved):

mogul (approved):

mogul:
Since Nick's not yet in the datagov GitHub team, GitHub Actions are wedged. James is going to submit a new PR with this content.

PR #318, opened 2021-06-11
Bugfix/dir listing

nickumia-reisys:
Possible fix for Directory Listing Issue.  I don't particularly like it because it overwrites the entire default Jetty configuration file.  If there is an update to that file, there's no easy way to compare and choose what changes we want to make.  The file is installed with solr in the base image.  The only reason I like this solution is because it sets the default access control for the entire web server to disabled as opposed to explicitly disabling this capability per server module.  Suggest better solutions if you see anything.

Please look at my commits for the change comments.  The most informative commit is, https://github.com/GSA/catalog.data.gov/pull/318/commits/7171d1db3e927aaa84f420116e11a3ab615d6d8f


FuhuXia:
Two suggestions:
1. Add comment, link to the github issue, where you made change in the Dockerfile, so we know what the change is for.
2. Instead of overwrite the whole webdefault.xml file, can you detect the dirAllowed block and change the block only? It is safer than replacing whole file.

nickumia-reisys:
Implemented suggestion and added appropriate comment. 👍

mogul (approved):
We're holding our noses on using perl, but the code is fine for what it does. ;)

PR #320, opened 2021-06-14
Bugfix/solr libraries update

nickumia-reisys:
Updated libraries for both https://github.com/GSA/datagov-deploy/issues/3284 and https://github.com/GSA/datagov-deploy/issues/3285.. While not ideal, this solution was taken for minimal impact to overall solr operation.  All vulnerabilities were referenced and online cdn's were referenced to download updated files.  Solr was using `cdnjs.cloudflare.com/ajax/libs` for some resources, so I used that link to procure additional files.

Additionally, I switched the solr version to 6.6.6-slim for greater portability amid memory and storage issues.  @adborden attempted to use this version originally since solr was previously on 6.2-alpine (https://github.com/GSA/catalog.data.gov/pull/298).  I ran all tests and cross-referenced the "Image Variants" on https://hub.docker.com/_/solr.  Unless there is some obvious reason we need the common packages, I think it's safe to go with the slim version.  This can be easily reverted before the pull request is merged.

### DIFF
--- a/solr/Dockerfile
+++ b/solr/Dockerfile
@@ -1,4 +1,4 @@
-FROM solr:6.6.6
+FROM solr:6.6.6-slim
 MAINTAINER Open Knowledge
 
 # Enviroment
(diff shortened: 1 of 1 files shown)


FuhuXia:
Anything out of order on the admin ui functionalities with the new versions of jquery and angularjs?

nickumia-reisys:
Actually, I need to re-test because it seems like there were permission errors that I didn't see before.  Right now, the links to the sub-pages don't work.  I will update soon.

nickumia-reisys:
@FuhuXia So.. it seems like there's a link resolution error where it tries to go to things like `http://localhost:8983/solr/#!/~java-properties` but it goes to `http://localhost:8983/solr/#!/#%2F~java-properties` instead.. The real page exists, but the link is broken  it adds an extra `%2F` that it doesn't resolve to a `/`  :( Other than that, I didn't find any other issues.

nickumia-reisys:
Documented the Solr Admin UI issues with the help of @jbrown-xentity and updated library overwrite approach per @FuhuXia's suggestion.  This is a pretty intense series of solr patches.. hopefully there aren't much more 😅

jbrown-xentity:
Are we able to confirm that we are using the newly downloaded version of angular? I don't see removal/deprecation of the older version of angular... @nickumia-reisys

jbrown-xentity:
> I didn't remove the older version, I do think that's the best thing to do. The verification is the update of the html scripts (sed replacement) to use the 3.6.0 version. Do we agree to explicitly remove the old file?

Replacing the pointer for jquery makes sense, and keeping the old files is fine (minimal edits to the docker image in that case). For angular, I'm not sure how the old files existed vs the new download. I see the files added, but not any removal of previous files or any change in pointing to the new files. Is this overwriting the old files? Something else?

nickumia-reisys:
Ohh... yes, In terms of the html code, it is overwriting the old files. Is this okay?

FuhuXia:
This is is replaced by ⬆️

PR #323, opened 2021-07-15
test harvesting and datajson collection

FuhuXia:
Added a data.json harvest source and tested its harvesting.
Verified collections in data.json.

For story https://github.com/GSA/datagov-deploy/issues/3325


jbrown-xentity (dismissed):
This all looks great! I thought we would do this at the same time as we upgrade the ckanext-datajson extension so that we can validate this use case, can we do an upgrade as well?

ckan/freeze-requirements.sh
```diff
@@ -6,6 +6,8 @@
 #
 # <user id> and <group id> are passed to make sure Pipfile.lock is owned by the correct user!
 
+set -e
```
nickumia-reisys, line 9: Was this just introduced to add additional error checking?

requirements/pyproject.toml
```diff
@@ -13,7 +13,7 @@ ckanext-archiver = {git = "https://github.com/ckan/ckanext-archiver.git"}
 ckanext-datagovtheme = {git = "https://github.com/GSA/ckanext-datagovtheme.git", branch = "main" }
 ckanext-datajson = {git = "https://github.com/GSA/ckanext-datajson.git", branch = "main" }
 ckanext-envvars = "*"
-ckanext-geodatagov = {git = "https://github.com/GSA/ckanext-geodatagov.git" }
+ckanext-geodatagov = {git = "https://github.com/GSA/ckanext-geodatagov.git", branch = "main" }
```
nickumia-reisys, line 16: Are we tracking the 'main' branch now because it's trusted?

tests/test-base.bats
```diff
@@ -51,60 +51,6 @@ load test_helper
   api_delete_call "organization" "purge" "test-organization-$RNDCODE"
 }
 
-@test "Given an organization, a waf-collection harvest source is created successfully from form" {
```
nickumia-reisys, line 54: Is this a depreciated test that is somewhat covered by the new test?

tests/test-datajson.bats
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env batsPI
+
+load globals
+load test_helper
+
+@test "datajson_harvest extension is loaded" {
+  test_extension_loaded datajson_harvest
+}
+
+@test "data.json harvest source is created and harvested successfully" {
```
nickumia-reisys, line 10: This seems like a pretty big test, are there any parts that can be split out into reusable functions for future tests?  Or is it all pretty specific to testing this feature?

ckan/freeze-requirements.sh
nickumia-reisys: Make sure error if poetry fails.

requirements/pyproject.toml
nickumia-reisys: because 'master' name change to 'main'

tests/test-base.bats
nickumia-reisys: the new test overwrites this functionality.

jbrown-xentity (approved):

tests/test-datajson.bats
FuhuXia: It is long test because the final UI test is depending the datasets harvested from previous actions. I agree the harvesting part can be separated into another function, but it is used only once. If other tests also use harvested datasets, then the DRY principle will kick in.

PR #324, opened 2021-07-20
fix poetry error due to geodatagov master->main change; make sure it fails on error

FuhuXia:
catalog is not pulling [latest xss fix](https://github.com/GSA/ckanext-datagovtheme/pull/90) from datagovtheme due to the poetry error.

same as commit 719a07b


FuhuXia:
deployed successfully on sandbox https://ci.sandbox.datagov.us/job/deploy-ci-platform-mb/job/develop/134/

XSS fix verified on sandbox calalog.

jbrown-xentity (approved):

PR #340, opened 2021-09-21
Feature/test harvest private

jbrown-xentity:
This should evaluate if a harvest source is created to be public that the config "stays", and when revisiting the edit page the public setting is still set.

This depends on https://github.com/GSA/ckanext-geodatagov/pull/193

The tests before ^^ changes are brought in should fail.

This is related to https://github.com/GSA/datagov-deploy/issues/897

### DIFF
--- a/e2e/cypress/support/command.js
+++ b/e2e/cypress/support/command.js
@@ -123,20 +123,15 @@ Cypress.Commands.add('delete_dataset', (datasetName) => {
 })
 
 
-Cypress.Commands.add('create_harvest_source', (dataSourceUrl, harvestTitle, harvestDesc, harvestType, org, harvestTest, invalidTest) => {
+Cypress.Commands.add('create_harvest_source', (dataSourceUrl, harvestTitle, harvestDesc, harvestType, harvestPrivate, invalidTest) => {
     /**
      * Method to create a new CKAN harvest source via the CKAN harvest form
      * :PARAM dataSourceUrl String: URL to source the data that will be harvested
      * :PARAM harvestTitle String: Title of the organization's harvest
      * :PARAM harvestDesc String: Description of the harvest being created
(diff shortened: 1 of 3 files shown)


e2e/cypress/support/command.js
```diff
@@ -123,20 +123,15 @@ Cypress.Commands.add('delete_dataset', (datasetName) => {
...
-Cypress.Commands.add('create_harvest_source', (dataSourceUrl, harvestTitle, harvestDesc, harvestType, org, harvestTest, invalidTest) => {
+Cypress.Commands.add('create_harvest_source', (dataSourceUrl, harvestTitle, harvestDesc, harvestType, harvestPrivate, invalidTest) => {
     /**
      * Method to create a new CKAN harvest source via the CKAN harvest form
      * :PARAM dataSourceUrl String: URL to source the data that will be harvested
      * :PARAM harvestTitle String: Title of the organization's harvest
      * :PARAM harvestDesc String: Description of the harvest being created
      * :PARAM harvestType String: Harvest source type. Ex: waf, datajson
-     * :PARAM org String: Organization that is creating the harvest source
-     * :PARAM harvestTest Boolean: Determines if to use UI in harvest source creation test or to follow the UI to create a source
```
nickumia-reisys, line 134: We should probably document, 
`:PARAM harvestPrivate String: Indicates whether the harvest is private ("True") or public ("False")`

nickumia-reisys (approved):
I like the that extension upgrade was the last commit so the tests fail in the second to last commit and upgrading the extension with the new stuff makes the tests pass. 🪀 🛷 🐶 (translation: you sly dog) 👍 

After mulling over the tests, I think they capture the existing functionality.  It doesn't look like any of the harvests are set to private, but I'm not sure if that's an important test.

jbrown-xentity:
99% of the time, harvest sources are public and we want them to stay public. This validates that use case.

PR #348, opened 2021-09-27
Set environment for docker or local

jbrown-xentity:
This works for inventory, but has not worked for catalog. Not sure why.


jbrown-xentity:
Did this to validate it wasn't a local machine issue. Closing.

PR #349, opened 2021-09-29
Update solr schema

jbrown-xentity:
In review of previous CKAN2.9 started work, it was found that some changes to SOLR were made. This pulls in those minor changes.


nickumia-reisys (approved):
I verified that it rebuilt the solr image and the tests pass. Apart from that, I don't know much about the actual changes.

PR #350, opened 2021-09-30
Update ckanext-harvest

jbrown-xentity:
Also bumped other things, including cryptography to 35.
That's a typo on the release, should be 3.5.0
https://github.com/pyca/cryptography/issues/6345

This uses a protected branch that we can merge other fixes into. Currently have applied 2 fixes, both of which having matching PR's waiting upstream that are waiting to be merged...

- https://github.com/GSA/ckanext-harvest/pull/20
- https://github.com/GSA/ckanext-harvest/pull/21

### DIFF
--- a/ckan/requirements.txt
+++ b/ckan/requirements.txt
@@ -14,11 +14,11 @@ ckanext-dcat @ git+https://github.com/ckan/ckanext-dcat@c4e7319aff55e5315831b28e
 ckanext-envvars @ git+https://github.com/GSA/ckanext-envvars.git@33f7e190ab332244cb961a425e09af592d9b647b
 -e git+https://github.com/GSA/ckanext-geodatagov.git@bc700f671cd1d890db265c65f5db1cba121297f4#egg=ckanext_geodatagov
 ckanext-googleanalyticsbasic @ git+https://github.com/GSA/ckanext-googleanalyticsbasic.git@c6a425d5e14d658c0fa3661fdc4423162161c3f4
--e git+https://github.com/GSA/ckanext-harvest.git@fd5a6b739666a62cff486873877eef10523aec1a#egg=ckanext_harvest
+-e git+https://github.com/GSA/ckanext-harvest.git@b653113abc059f58917732f33a3ae2c84a476746#egg=ckanext_harvest
 -e git+https://github.com/gsa/ckanext-spatial.git@36981c39d624685bc85b84f801f2a2d6eaea67f6#egg=ckanext_spatial
 ckantoolkit==0.0.3
 click==7.1.2
-cryptography==3.4.8
+cryptography==35.0.0
 defusedxml==0.7.1
 distro==1.6.0
 dominate==2.4.0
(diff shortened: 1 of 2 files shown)


ckan/requirements.txt
```diff
@@ -14,11 +14,11 @@ ckanext-dcat @ git+https://github.com/ckan/ckanext-dcat@c4e7319aff55e5315831b28e
 ckanext-envvars @ git+https://github.com/GSA/ckanext-envvars.git@33f7e190ab332244cb961a425e09af592d9b647b
 -e git+https://github.com/GSA/ckanext-geodatagov.git@bc700f671cd1d890db265c65f5db1cba121297f4#egg=ckanext_geodatagov
 ckanext-googleanalyticsbasic @ git+https://github.com/GSA/ckanext-googleanalyticsbasic.git@c6a425d5e14d658c0fa3661fdc4423162161c3f4
--e git+https://github.com/GSA/ckanext-harvest.git@fd5a6b739666a62cff486873877eef10523aec1a#egg=ckanext_harvest
+-e git+https://github.com/GSA/ckanext-harvest.git@b653113abc059f58917732f33a3ae2c84a476746#egg=ckanext_harvest
 -e git+https://github.com/gsa/ckanext-spatial.git@36981c39d624685bc85b84f801f2a2d6eaea67f6#egg=ckanext_spatial
 ckantoolkit==0.0.3
 click==7.1.2
-cryptography==3.4.8
+cryptography==35.0.0
```
nickumia-reisys, line 21:
```suggestion
cryptography==3.5.0
```

ckan/requirements.txt
```diff
@@ -49,7 +49,7 @@ Mako==1.1.5
 Markdown==3.1.1
 MarkupSafe==2.0.1
 messytables==0.15.2
-newrelic==6.10.0.165
+newrelic==7.0.0.166
```
nickumia-reisys, line 52: It's the same increment on the minor release, even with the point release upgrade?

nickumia-reisys (dismissed):
This will be messy going forward 😕

ckan/requirements.txt
nickumia-reisys: Oh, I saw the link now.. that's a pretty big mess up haha 😓

jbrown-xentity:
So the tests fail because it fails cleanup. The CSW and WAF harvest sources are having the same datajson issue with the datasets not being properly assigned with the harvest source. When a clear is run, the datasets aren't affected in solr in the whole thing fails. Not sure if this is introduced by us or upstream, investigating...

ckan/requirements.txt
jbrown-xentity: Until it's registered on pypi, we have to use the published version... https://pypi.org/project/cryptography/#history

ckan/requirements.txt
jbrown-xentity: Yep, it's a weird way to track builds... https://pypi.org/project/newrelic/#history

PR #353, opened 2021-10-01
Bug/fix catalog map search

NeerajaVadlapudi:
- Reverted datagovtheme commit id from  [#29d745](https://github.com/GSA/ckanext-datagovtheme/commit/29d7454afd516b63cdabb41e7b1dfd4f69daffd4) to [#704c45](https://github.com/GSA/ckanext-datagovtheme/commit/704c45921a9246fd6072ea99711992e57d3bc416) to load the map properly
- Issue is due to change in the url method name from **organization.index** to **organizations_index**


FuhuXia:
Good findings to narrow down the commit to dependency `datagovtheme` where it creates the bug. Since both `main` and `fcs` branches have map feature broken, we dont want to simple pin it on `fcs` branch and leave `main` branch hanging. Can we fix it at the root, ie, in repo `datagovtheme`?

nickumia-reisys:
(On FCS) I reverted this change, rebuilt the ckan image, verified that datagovtheme was at the older commit and the map still wasn't loading for me.  @NeerajaVadlapudi Do you want to make a test that the map loads?

NeerajaVadlapudi:
I tested with reverted change i am able to see the map in my local please find below is the attached image for your reference (i tested this on fcs branch)

Map loading

PR #356, opened 2021-10-06
Upgrade ckanext-spatial

jbrown-xentity:
Test that a spatial harvested dataset with bad tags is still editable,
and can be added to a group.

Related to https://github.com/GSA/ckanext-spatial/pull/12 & https://github.com/GSA/datagov-deploy/issues/2413

### DIFF
--- a/e2e/cypress/integration/spatial.spec.js
+++ b/e2e/cypress/integration/spatial.spec.js
@@ -1,0 +1,30 @@
+describe('Spatial', () => {
+    it('Can search geographies by name', () => {
+        cy.request('/api/3/action/location_search?q=california').should((response) => {
+            expect(response.body).to.have.property('success', true);
+            expect(response.body.result[0]).to.have.property('text', 'California');
+        });
+    });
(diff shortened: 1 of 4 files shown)


e2e/cypress/integration/spatial.spec.js
```diff
@@ -0,0 +1,30 @@
+describe('Spatial', () => {
+    it('Can search geographies by name', () => {
+        cy.request('/api/3/action/location_search?q=california').should((response) => {
+            expect(response.body).to.have.property('success', true);
+            expect(response.body.result[0]).to.have.property('text', 'California');
+        });
+    });
+    it('Can put a package with weird tags in an group', () => {
```
nickumia-reisys, line 8: What about this introduces weird tags?

nickumia-reisys: The `>` signs?

jbrown-xentity: That, but also the `/` and the spaces...

PR #358, opened 2021-10-08
Add java and saxon to Docker

jbrown-xentity:
Related to https://github.com/GSA/datagov-deploy/issues/2917
Install java & saxon via apk.
Run CSDGM/FGDC metadata harvest test.
cloud.gov installation/setup for java.

Deployment [tested on develop](https://github.com/GSA/catalog.data.gov/runs/3944957334?check_suite_focus=true).
Catalog harvesters aren't [running yet in cloud.gov](https://github.com/GSA/datagov-deploy/issues/2810), but a manual test was run via ssh (`cf ssh catalog -t -c "/tmp/lifecycle/launcher /home/vcap/app bash ''"`):

```
vcap@6f46d7d3-8593-4805-49ab-d2ca:~$ java --version
openjdk 11.0.11 2021-04-20
OpenJDK Runtime Environment (build 11.0.11+9-Ubuntu-0ubuntu2.18.04)
OpenJDK 64-Bit Server VM (build 11.0.11+9-Ubuntu-0ubuntu2.18.04, mixed mode)
vcap@6f46d7d3-8593-4805-49ab-d2ca:~$ java net.sf.saxon.Transform -s:/home/vcap/deps/1/src/ckanext-geodatagov/ckanext/geodatagov/tests/data-samples/waf-fgdc/fgdc-csdgm_sample.xml -xsl:/home/vcap/deps/1/src/ckanext-geodatagov/ckanext/geodatagov/harvesters/fgdcrse2iso19115-2.xslt
<?xml version="1.0" encoding="UTF-8"?>
<!--+########################################################################+-->
<!--+    This file was generated using an FGDC CSDGM to ISO19139 Transform   +-->
<!--+                                                                        +-->
<!--+ This is accomplished using an XML Style sheet that transforms metadata +-->
<!--+ conforming to the CSDGM to the ISO 19139.                              +-->
<!--#########################################################################+-->
<gmd:MD_Metadata xmlns:gco="http://www.isotc211.org/2005/gco"
                 xmlns:gmd="http://www.isotc211.org/2005/gmd"
                 xmlns:gmi="http://www.isotc211.org/2005/gmi"
                 xmlns:gmx="http://www.isotc211.org/2005/gmx"
                 xmlns:gsr="http://www.isotc211.org/2005/gsr"
                 xmlns:gss="http://www.isotc211.org/2005/gss"
                 xmlns:gts="http://www.isotc211.org/2005/gts"
                 xmlns:gml="http://www.opengis.net/gml/3.2"
                 xmlns:xlink="http://www.w3.org/1999/xlink"
                 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                 xsi:schemaLocation="http://www.isotc211.org/2005/gmd  http://schemas.opengis.net/iso/19139/20070417/gmd/gmd.xsd">
...
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,3 @@ e2e/cypress/screenshots/*
 e2e/cypress/results/output.xml
 node_modules
 package-lock.json
-cflinuxfs3-fhs-fixup/.bash_history
-cflinuxfs3-fhs-fixup/.lesshst
-cflinuxfs3-fhs-fixup/*.deb


cfsetup.sh
```diff
@@ -2,7 +2,7 @@
 
 set -o errexit
 set -o pipefail
-
+# TODO: convert to .profile
```
robert-bryson, line 5: Is this TODO done by this [this new .profile](https://github.com/GSA/catalog.data.gov/blob/68203482643f9b45eda23157f6f956482a1e890f/.profile)/.

jbrown-xentity: Good question: no. We discovered we could use a .profile file to do basically what cfsetup.sh is doing in a custom way, and thought we should do it the "normal" way. See Bret's comments [here](https://github.com/GSA/datagov-deploy/issues/3251#issuecomment-946103785).
The reason I left a TODO instead of doing the work is that I noticed cfsetup.sh uses bash, but .profile uses shell. I decided it was a separate task to merge those together, and I wasn't familiar with what could be done (can we make .profile a bash script? What are the ramifications of this? What's best practice?). I'm going to make a comment in that dashboard ticket requesting to look at this when the ticket is taken on....

robert-bryson (approved):

PR #359, opened 2021-10-12
Pin geodatagov

jbrown-xentity:
Pin FCS branch with old geodatagov version, so that it can continue to work with the FGDC2ISO external transformation service.

Related to https://github.com/GSA/datagov-deploy/issues/2917


nickumia-reisys:
The updates aren't backwards compatible?  It seems like a waste that we make things work on 2.8 and 2.9 if they don't actually work in 2.8 😕

robert-bryson (approved):

PR #363, opened 2021-10-21
Update ckanext-datagovtheme [main]

jbrown-xentity:
We are fixing a bug that cannot be tested downstream with python tests, is a UI fix.
Added a cypress test to simulate clicking on the map button, and draw a bounding box to search in. This will suffice to confirm this type of issue does not arise in the future.
Related to https://github.com/GSA/datagov-deploy/issues/3434

Also removes an old test file from the bats system, no longer needed.


jbrown-xentity:
Removal of Dockerfile.test is just an old artifact that should have been removed on moving to cypress tests...

nickumia-reisys (approved):

PR #370, opened 2021-11-02
Add SOLR as a service

jbrown-xentity:
Related to https://github.com/GSA/datagov-deploy/issues/2778

Updates services to be created if they don't exist
Adds SOLR as a service, removes it as a different app.

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -125,7 +125,7 @@ jobs:
         if: steps.check-catalog-solr-dev.outcome == 'failure'
         uses: usds/cloud-gov-cli@master
         with:
-          command: create-service solr-cloud base catalog-solr -c '{"solrImageRepo":"ghcr.io/gsa/catalog.data.gov.solr", "solrImageTag":"latest"}' -b ssb-solr-gsa-datagov-development --wait
+          command: create-service solr-cloud base catalog-solr -c solr/service-config.json -b ssb-solr-gsa-datagov-development --wait
           org: gsa-datagov
           space: development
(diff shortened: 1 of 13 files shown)


jbrown-xentity:
This PR is almost ready to go. All that should need to happen is specify what parameters are needed on catalog-solr service, and then test creating it via the same process on the develop branch. Will need to change the start script to remove the environment-passed URL to the solr service, and parse from the vcap services...

mogul:
I rebased on main, added some documentation to the `README.md`, and force-pushed. However, I haven't had a chance to try it out or update the GitHub Actions to create the service yet.

adborden:
Why are we creating the services in GitHub Actions? These are essentially one-off commands since the service lives through the lifetime of the application. A one-liner in the docs under required services isn't enough?

mogul:
Two reasons: The first is that in a Contingency Planning situation we want recovery to require  as few steps as possible. The second is that we want as few steps as possible between forking the repository and having a live deployment, which puts us one step closer to making it trivial for others to retrace our steps.

robert-bryson (approved):

mogul:
Now using the SolrCloud collection setup script

mogul:
Switching back to draft status on this one until we refactor the service creation logic to hoist it out of the GitHub Actions and make it DRY.

mogul:
BEFORE MERGING: Delete the existing `catalog-*` services in `staging` and `prod`, so that the merge will recreate them with appropriately-sized plans.

robert-bryson (dismissed):

robert-bryson (approved):

PR #371, opened 2021-11-17
add saml2auth, update requirements

FuhuXia:
For https://github.com/GSA/datagov-deploy/issues/3508

It has been deployed to [develop environment](https://catalog-dev-datagov.app.cloud.gov/).


jbrown-xentity (approved):

jbrown-xentity:
Confirmed working! Do we want to set this up in dev as well, are we able to get that working with tests?
image

FuhuXia:
> Confirmed working! Do we want to set this up in dev as well, are we able to get that working with tests?
We should, since I observed some UI glitches with saml2auth login turned on. Not sure how thou. Can't use login.gov. Maybe have to use a local IdP.

PR #374, opened 2021-11-22
Change processes to separate apps

jbrown-xentity:
Change processes to be separate apps, and manage accordingly.
Setup restarts and deployments for each.

Deployment changes worked in dev, see [here](https://github.com/GSA/catalog.data.gov/runs/4291290963?check_suite_focus=true).


robert-bryson (approved):

jbrown-xentity:
The following error occurred: `JQuery(...).on is not a function`, on the resource page of a dataset. Re-trying test, but if it re-occurs then something is wrong with the resource pages...

PR #395, opened 2021-12-30
Download cypress results from github actions

nickumia-reisys:
Attempt to follow [Artifact Download Instructions](https://github.com/actions/download-artifact) to see the results of failed cypress tests on Github Actions.  This may need to be in the context of the action that uploaded the artifacts.  But based on the documentation, it doesn't seem like it has to be.

Note: artifacts are case-insensitive and will overwrite each other.  Our current cypress tests don't name each run, so there's probably only one set of artifacts available to download (which would be the most recent run).

### DIFF
--- a/.github/workflows/download-cypress.yml
+++ b/.github/workflows/download-cypress.yml
@@ -1,0 +1,21 @@
+
+name: Download Cypress Results
+on:
+  workflow_dispatch:
+
+jobs:
+  download:
+    name: 'Download'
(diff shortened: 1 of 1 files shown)


nickumia-reisys:
I don't know if this will work.  But I can't test it running until it's on `main`.

nickumia-reisys:
Per James' insight, files are available for download on the bottom of the "summary" page in the Github Action rule.

PR #403, opened 2022-01-13
Implement NGINX Proxy

nickumia-reisys:
Related to https://github.com/GSA/datagov-deploy/issues/3629


nickumia-reisys:
Super strange bug... I was able to add the network policy once, but then after that I kept getting errors...
image

image

nickumia-reisys:
Note: it was working in staging.. but then it stopped.  When it was working, I saw requests like,

image


But then, I think the network-policy got messed up, maybe because of a restart? And then it started giving errors like,

image


It looks like it's still serving traffic, but the proxy is not returning the traffic to the outside world 😕

mogul:
That's true, I only put dev versions in the dev space, triggered by hand, when needed. (We mostly develop the broker locally.) The process is the same as for prod and staffing; I can show you how to do it.

Regarding the intermittent support for the network policy API: I saw that a few times in the last couple days as well! I found that logging out at the command line and back in resolved it, though I can't figure out why that would do it.

nickumia-reisys:
The only reason I was going to deploy the broker in development was to test catalog app in development space which would require the solr service be created.  It's a mute point now since we're using staging to test the proxy.

proxy/nginx.conf
```diff
@@ -0,0 +1,74 @@
...
+    root public;
+    index index.html index.htm Default.htm;
+  }
+
+  ## Gunicorn specs
+  server {
+    # use 'listen 80 deferred;' for Linux
+    # use 'listen 80 accept_filter=httpready;' for FreeBSD
+    listen {{port}};
+    client_max_body_size 4G;
```
FuhuXia, line 29: The default 4G from gunicorn  is awfully big for catalog. Can we start with 250m?

proxy/public/500.html
```diff
@@ -0,0 +1,2 @@
+<h1>500 Web server unavaiable 😵</h1>
+<p>Please contact datagovhelp@gsa.gov for assistance.</p>
```
FuhuXia, line 2: On existing FCS config, We have custom static page for 500 error hosted on filestore.data.gov. We can reuse it. Maybe a separated ticket after the clould.gov move.

FuhuXia:
During the verification on stage, we noticed 500 error if catalog is manually restarted, and the 500 error persists until catalog-proxy is manually restarted. It wont be a big issue since the scheduled restart action always restart both apps, but in theory we might see the 500 error for a split second (or longer) once in a while when things are restarting.

mogul:
Does that happen even if you do `cf restart catalog --strategy rolling`? It shouldn't if so!

proxy/nginx.conf
nickumia-reisys: Done.

nickumia-reisys:
The last item of concern (raised by @FuhuXia) is surrounding SEO web crawling and internet identification of catalog.data.gov.  If the backend app is indexed in isolation from the proxy app, it will confuse people as to why there are two different sites showing up on google for catalog.data.gov.

Are we accepting this risk until we move to cloud.gov?

mogul:
I think it's reasonable to accept, but it's also reasonable to fix! Add a robots.txt to the backend site, and don't proxy just that one file in nginx.

FuhuXia (approved):

FuhuXia:
> Follow up note: ckan implements a robots.txt, but I don't know if it does what we would want it to do.. https://github.com/ckan/ckan/blob/master/ckan/public/robots.txt

What we can do is keeping this robots.txt as it is, but config nginx to serve a new robots.txt when the host is the internal domain:
```
User-agent: *
Disallow: /
```

PR #404, opened 2022-01-13
Fix Harvester Service Retrievals

nickumia-reisys:
The harvesters are run is separate apps, but they use the same services as the main app, so then they lookup service names, they need to use 'catalog' as the APP_NAME, not 'catalog-gather' or 'catalog-fetch'


jbrown-xentity (approved):

jbrown-xentity:
I can't believe I didn't think of this... Good catch @nickumia-reisys !

PR #408, opened 2022-01-19
Update site url appropriately

jbrown-xentity:
CKAN needs the site url to be set to the correct domain in order to serve responses properly. This is more complicated, and we need to be more explicit now that there is an nginx proxy.


nickumia-reisys (approved):
Teamwork makes the dream work 🎆

PR #411, opened 2022-02-02
ckan 2.8.10

FuhuXia:


nickumia-reisys (approved):
Don't know much about FCS compatibility for upgrade, but approving to move along 👾

robert-bryson (approved):

PR #421, opened 2022-02-11
bring in map fix from datagovtheme

FuhuXia:
For https://github.com/GSA/datagov-deploy/issues/3587.

PR #422, opened 2022-02-14
bring in changes from geodatagov and datagovtheme for tag fix

FuhuXia:
For https://github.com/GSA/datagov-deploy/issues/3630.


jbrown-xentity (approved):

jbrown-xentity:
@FuhuXia looks like it failed, you can download the video of the error [here](https://github.com/GSA/catalog.data.gov/actions/runs/1842980610). It definitely has to do with the spatial change, it's trying to grab the `.leaflet-control-draw-rectangle` item but it can't be found; it might be called something different in upstream?

FuhuXia:
> @FuhuXia looks like it failed, you can download the video of the error [here](https://github.com/GSA/catalog.data.gov/actions/runs/1842980610). It definitely has to do with the spatial change, it's trying to grab the `.leaflet-control-draw-rectangle` item but it can't be found; it might be called something different in upstream?

The error has been fixed in https://github.com/GSA/catalog.data.gov/pull/421. Waiting for that PR to be merged then we can update this one.

nickumia-reisys (approved):
🎸

PR #423, opened 2022-02-17
updated to use nginx proxy

robert-bryson:
Related to https://github.com/gsa/datagov-deploy/issues/3629


jbrown-xentity:
We were unable to get nginx running as a sidecar or in a multi-buildpack; were able to move in a different direction. See #427.

PR #427, opened 2022-03-07
Feature/internal routes

robert-bryson:
Related to https://github.com/gsa/datagov-deploy/issues/3629

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -4,7 +4,7 @@ services:
   test:
     image: cypress/included:7.3.0
     entrypoint: []
-    command: /bin/bash -c "npx wait-on http://ckan:5000 && cypress run"
+    command: /bin/bash -c "npx wait-on http://ckan:5000/dataset && cypress run"
     # To run specific test, use the following syntax:
     # command: /bin/bash -c "npx wait-on http://ckan:5000 && cypress run --spec cypress/integration/login.spec.js"
     environment:
--- a/ckan/requirements.in
+++ b/ckan/requirements.in
@@ -97,10 +97,10 @@ click==7.1.2
 dominate==2.4.0
 fanstatic==1.1
 feedgen==0.9.0
-Flask==1.1.1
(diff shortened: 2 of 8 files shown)


jbrown-xentity (dismissed):

docker-compose.test.yml
```diff
@@ -4,7 +4,7 @@ services:
   test:
     image: cypress/included:7.3.0
     entrypoint: []
-    command: /bin/bash -c "npx wait-on http://ckan:5000 && cypress run"
+    command: /bin/bash -c "npx wait-on http://ckan:5000/dataset && cypress run"
```
nickumia-reisys, line 7: This shouldn't matter.  The test work locally for me without this change.  I think this might just be another blip in catalog tests unfortunately 😕

jbrown-xentity: Bringing it up locally, one of the requirements bumps (jinja2 dependencies being loose) causes the app to fail to load. @robert-bryson will push a fixed version... We thought maybe npx didn't like the 302 that occurs when hitting root and that hitting /dataset would work cleaner, but it was the requirement bump that was the issue.

nickumia-reisys: Interesting haha.. I didn't rebuild, so yeah, that makes sense 😅

jbrown-xentity (approved):
Looks great!

robert-bryson:
Remind to self: after build, need to keep an eye on the network policies to make sure the proxy can communicate with catalog in the stage/prod spaces. Not sure if it will be magically added. 

Should look like this:
```
➜ cf network-policies                                                                                  
Listing network policies in org gsa-datagov / space development as robert.bryson@gsa.gov...

source          destination   protocol   ports   destination space   destination org
catalog-proxy   catalog       tcp        61443   development         gsa-datagov
```

If it isn't there, can be added with: `cf add-network-policy catalog-proxy catalog --protocol tcp --port 61443`.

PR #439, opened 2022-03-28
Fix github action syntax for Harvester Jobs

nickumia-reisys:


nickumia-reisys:
Successful run: https://github.com/GSA/catalog.data.gov/runs/5725681813?check_suite_focus=true

image

.github/workflows/run-harvest-check.yml
```diff
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 jobs:
-#   # TODO: delete when reenabled
+#   # Note: run this action if the harvesters need to be commented out
```
jbrown-xentity, line 10: Why would the harvesters need to be commented out?

nickumia-reisys: If we ever want to "turn off the harvesters" again just like we needed to for the transition.  I'm not foreseeing anything specifically, just mentioning.

FuhuXia (approved):

PR #440, opened 2022-03-30
Pause harvesters for solr switch

nickumia-reisys:


nickumia-reisys:
In order to stop harvesters, we can just delete the `catalog-gather` and `catalog-fetch` apps.  Even if the `ckan harvester run` command runs in the main web app, it won't be able to do any meaningful work.

FuhuXia (approved):

PR #445, opened 2022-04-08
restart every 30 mins

FuhuXia:
For https://github.com/GSA/data.gov/issues/3664.
Why 15 if we can get by with 30 mins.


nickumia-reisys:
I thought this was a cyber requirement to restart every 15 mins?

nickumia-reisys (approved):
⏲️

PR #455, opened 2022-04-29
Revert: solr_commit = false

nickumia-reisys:
No comment. 🌊

### DIFF
--- a/ckan/setup/ckan.ini
+++ b/ckan/setup/ckan.ini
@@ -101,8 +101,8 @@ ckan.auth.create_default_api_keys = false
 ## Search Settings
 
 ckan.site_id = datagov_catalog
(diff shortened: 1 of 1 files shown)


jbrown-xentity (approved):
Is there a reason we're turning this back off? Was this utilized from a previous test case?

FuhuXia:
https://docs.ckan.org/en/2.9/maintaining/configuration.html#ckan-search-solr-commit
The previous change was done in an attempt to reduce solr load on our solrcloud debugging. It turned out not necessary. In order to make it work, additional config is needed to be done on solr side to enable soft commit, otherwise solr will not show any change on document count until we reload the core/collection. 

Turn it back to default True is an easier thing to do.

PR #460, opened 2022-05-27
Use standalone Solr on ECS

nickumia-reisys:
Related to https://github.com/GSA/data.gov/issues/3826

New additions/changes:
- Build Solr Docker Images with `efs-utils`, `vim` and other necessary tools
- Switch local docker environment to standalone solr from solrcloud
- Configure Solr on ECS config paramters in `service-config.json`

### DIFF
--- a/ckan/setup/GSA_prerun.py
+++ b/ckan/setup/GSA_prerun.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import time
+import requests
 try:
     from urllib.request import urlopen
     from urllib.error import URLError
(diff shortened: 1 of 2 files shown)


btylerburton (dismissed):

ckan/setup/GSA_prerun.py
```diff
@@ -50,5 +50,6 @@ def check_solr_connection(retry=None):
         pr.update_plugins()
         pr.check_datastore_db_connection()
         pr.init_datastore_db()
-        check_solr_connection()
+        # This function does not work, but solr is up
+        # check_solr_connection()
```
jbrown-xentity, line 54: I know we had to overwrite the upstream version of this function when we moved to solr8 and/or solr cloud, maybe it would work now?

jbrown-xentity (approved):
Looks great!

robert-bryson:
I, for one, appreciate that `vim` is a necessary tool.

ckan/setup/GSA_prerun.py
jbrown-xentity: Nope, this is cloud.gov deployment. As long as all the pieces are there (db, solr, etc) then this should just work.

btylerburton: Sweet. Will give this a shot then.

jbrown-xentity: Added solr back into GSA_prerun.py, removed some bash for solr setup.

FuhuXia (approved):

PR #466, opened 2022-06-07
Feature/collection fixes

jbrown-xentity:
Related to https://github.com/GSA/data.gov/issues/3834.

~Expect tests to fail, until we have a working version of ckanext-geodatagov~

UPDATE 7/9: All tests passing.


.github/workflows/commit.yml
```diff
@@ -11,28 +11,9 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v2
-      - name: Verify Changed files
```
robert-bryson, line 14: I'm not super familiar with what this is actually doing, but what's the reason for it to be removed?

btylerburton: James removed that in the pairing call since it wasn't actually lookin for changed files in the commit but rather files changes during the build IIRC.

btylerburton: I believe we would still want a changed files check but one that looks at the PR level.

robert-bryson (approved):

btylerburton:
TODO: add note to .ini file

PR #474, opened 2022-06-17
Update logging config

jbrown-xentity:
Pulled from various best practices of Python logging. This should make standard debug, info, and warning messages appear as normal in stdout (and will be handled appropriately/as such in various log systems). Only actual errors should appear in stderr.

See https://docs.python.org/3/howto/logging.html#configuring-logging and https://stackoverflow.com/questions/25187083/python-logging-to-multiple-handlers-at-different-log-levels

*Update*: development deploy looks good, normal log output is clean while errors are highlighted per what you would expect:
image

### DIFF
--- a/ckan/setup/ckan.ini
+++ b/ckan/setup/ckan.ini
@@ -286,49 +286,55 @@ ckanext-archiver.cache_url_root={{ckan_site_domain}}
 keys = root, ckan, ckanext, werkzeug, saml2, model
 
 [handlers]
-keys = console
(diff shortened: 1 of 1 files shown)


robert-bryson (approved):

jbrown-xentity:
@Jin-Sun-tts @btylerburton would either of you like to implement this on inventory?

btylerburton:
Glad to. Let me create an issue and connect it to this one.

PR #481, opened 2022-06-24
Remove duplicate cores from solr data

nickumia-reisys:
When an EFS restore happens, it causes a duplicate ckan core.  This first fix just deletes the restored core since we think the original core may still be functional

Related to https://github.com/GSA/data.gov/issues/3862 and https://github.com/GSA/data.gov/issues/3869


FuhuXia (approved):

FuhuXia:
💬

PR #492, opened 2022-07-11
enable ckan cache

FuhuXia:
Part of https://github.com/GSA/data.gov/issues/3887

This will make ckan to response with `cache-control: public` instead of `private`, so most pages will cached by CDN.

https://docs.ckan.org/en/2.9/maintaining/configuration.html#ckan-cache-enabled


ckan/setup/ckan.ini
```diff
@@ -42,6 +42,7 @@ beaker.session.timeout=900
 
 # CKAN caching
 ckan.cache_expires = 86400
```
jbrown-xentity, line 44: Caching pages for 24 hours seems agressive; do you think we could turn that down to maybe 1 hour?

jbrown-xentity (dismissed):

ckan/setup/ckan.ini
FuhuXia: 1 hour expiry will increase server load. Let us play safe and go with default value first. We can re-evaluate a good expiry after things are running on cloud.gov after a while.

FuhuXia: based on preliminary load tests on development space, it is safe to use 1 hour cache time instead of 1 day.

nickumia-reisys (approved):
🐐

PR #494, opened 2022-07-13
Adds new test for harvest source spatial format

btylerburton:
Adds new dataset to support the above
Swaps out CSW dataset to improve speed
Pins version of cypress interactive to same version in CLI
Cleans up tests before/after to not persist login across files
Updates dependencies and pins new geodatagov version


e2e/cypress.json
```diff
@@ -1,8 +1,4 @@
 {
     "baseUrl": "http://ckan:5000",
-    "reporter": "junit",
-    "reporterOptions": {
-        "mochaFile": "cypress/results/output.xml"
```
nickumia-reisys, line 5: Is there any reason we don't want this output file anymore?  If we're worried about not seeing results in the terminal, we could add the `toConsole` parameter as well: https://docs.cypress.io/guides/tooling/reporters#Cypress-configuration-2

e2e/cypress/integration/ckan_extensions.spec.js
```diff
@@ -23,14 +25,18 @@ describe('CKAN Extensions', () => {
             expect(installed_extensions).to.include('geodatagov_miscs');
             expect(installed_extensions).to.include('z3950_harvester');
             expect(installed_extensions).to.include('arcgis_harvester');
-            expect(installed_extensions).to.include('geodatagov_geoportal_harvester');
+            expect(installed_extensions).to.include(
+                'geodatagov_geoportal_harvester'
+            );
```
nickumia-reisys, line 30: Is there a line limit lint that put this on a new line?

e2e/cypress/integration/ckan_extensions.spec.js
```diff
@@ -23,14 +25,18 @@ describe('CKAN Extensions', () => {
...
             expect(installed_extensions).to.include('waf_harvester_collection');
             expect(installed_extensions).to.include('geodatagov_csw_harvester');
             expect(installed_extensions).to.include('geodatagov_doc_harvester');
             expect(installed_extensions).to.include('geodatagov_waf_harvester');
             expect(installed_extensions).to.include('spatial_metadata');
             expect(installed_extensions).to.include('spatial_query');
-            expect(installed_extensions).to.include('spatial_harvest_metadata_api');
+            expect(installed_extensions).to.include(
+                'spatial_harvest_metadata_api'
+            );
```
nickumia-reisys, line 39: Same as last comment, expanding this seems trivially bad haha..

nickumia-reisys (approved):
Lots of good linting 🗃️

nickumia-reisys (dismissed):
🗃️ 🗃️

e2e/cypress.json
btylerburton: Oooh, good suggestion. Will rerun locally with those options and see if we get what we want.

e2e/cypress/integration/ckan_extensions.spec.js
btylerburton: Yeah, this is Prettier's out of the box limit, which is set to 80 characters. I feel like I've seen a python limit somewhere, but can't recall where I've seen that. Do you know offhand?

nickumia-reisys: Yeah, the default python is also 80.  We set it to 128 in some of our repos, but yeah, I think somewhere around 120 should be a good limit.

e2e/cypress.json
nickumia-reisys: No, probably not.  So it's okay to leave it as you changed it.  Just wanted to mention 😅

btylerburton: No worries. Always glad to spend a bit of effort to figure out what we really want. 

FYI current output to static output.xml only includes the last spec -- by design. https://docs.cypress.io/guides/tooling/reporters#Merging-reports-across-spec-files

e2e/cypress/integration/ckan_extensions.spec.js
btylerburton: Changed my prettier config to 128 and repushing.

e2e/cypress.json
nickumia-reisys: Good point, it wasn't entirely working before anyway haha..

btylerburton:
just added prettierrc and had it lint the whole dir. @nickumia-reisys can i get another quick pass whenever it's convenient? thanks!

nickumia-reisys (approved):
🗃️ 🗃️ 🗃️

PR #500, opened 2022-07-26
Add new app catalog-admin

FuhuXia:
For https://github.com/GSA/data.gov/issues/3893

This PR adds new app catalog-admin.
In Nginx it is set to allow traffic from within GSA network to access this app. 
Same SAML app was used on identitysandbox.gov with SP changed to catalog-admin URL.

Post deployment
Add network policy to allow catalog-proxy access catalog-admin.


nickumia-reisys (approved):
📟

PR #511, opened 2022-08-03
Check login cookie

jbrown-xentity:
Don't force GSA network access for admin login

Related to https://github.com/GSA/data.gov/issues/2788

Login and logout work as expected in development, ready for review...

### DIFF
--- a/proxy/nginx.conf
+++ b/proxy/nginx.conf
@@ -96,12 +96,20 @@ http {
     set_real_ip_from 0.0.0.0/0;
     real_ip_header X-Forwarded-For;
     real_ip_recursive on;
-    set $allow false;
-    if ($remote_addr ~ "159.142.") {
-      set $allow true;
+    set $allow "";
+    # Determine if auth cookie is set
+    if ($cookie_auth_tkt) {
+      set $allow C;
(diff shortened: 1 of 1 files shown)


proxy/nginx.conf
```diff
@@ -91,12 +91,24 @@ http {
...
+      set $authy "${authy}S";
+    }
+    if ($uri = "/api/action/status_show") {
+      set $authy "${authy}H";
     }
-    if ($allow = false) {
-      return 403;
+    # If not logged in and not trying to login, redirect
+    if ($authy = "") {
+      return 302 https://{{env "EXTERNAL_ROUTE"}};
```
FuhuXia, line 111:
```suggestion
      return 302 https://{{env "PUBLIC_ROUTE"}};
```

FuhuXia: Let us use PUBLIC_ROUTE as much as possible for redirect. It is CDN if CDN is available, or same value as EXTERNAL_ROUTE if CDN not available.

FuhuXia (changes requested):

FuhuXia (approved):

PR #512, opened 2022-08-04
Nginx conf rework

FuhuXia:
Some refactoring some new features. Each commit should be atomic.

related to https://github.com/GSA/data.gov/issues/3909


jbrown-xentity (dismissed):
I really hated the duplication of code from before, but didn't want to take the time to dig into nginx and map docs to figure out how to fix it. This is much better!!! Nice work @FuhuXia

jbrown-xentity (approved):

PR #518, opened 2022-08-09
use simple solr locktype; add retry in case of lock

FuhuXia:
Related to https://github.com/GSA/data.gov/issues/3920

Use `simple` LockType and some retry logic to have a successful Solr Task restart without Core file being locked or Core file be corrupted.


nickumia-reisys (approved):
🧼

PR #529, opened 2022-08-24
Allow CORS for data.gov

nickumia-reisys:
Related to 
- https://github.com/GSA/data.gov/issues/3891
- https://github.com/GSA/inventory-app/pull/444

To support dataset count gathering, open catalog.data.gov to data.gov


jbrown-xentity (approved):
Do we want to whitelist staging and dev as well, so we can confirm these changes work as expected?

btylerburton (approved):

btylerburton:
James has a good point there.

PR #530, opened 2022-08-24
Update requirements to fix cloud.gov build

nickumia-reisys:
This fixes the recent build failure after some recent cflinuxfs3 release. Discussion in [cg-support slack channel](https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1661355526209379).


btylerburton (approved):
worth trying to see if it fixes our build, but i did see that it was advised against using the binary for production

https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary

jbrown-xentity (approved):

PR #531, opened 2022-08-24
Add CORS origin in nginx proxy

nickumia-reisys:
Fixes
- #529 

The CKAN configuration does not work, so add it to the Proxy instead.

### DIFF
--- a/ckan/setup/ckan.ini
+++ b/ckan/setup/ckan.ini
@@ -115,9 +115,7 @@ ckan.site_id = datagov_catalog
 
 ## CORS Settings
 
-# Allow dataset count to populate on data.gov
-ckan.cors.origin_allow_all = False
-ckan.cors.origin_whitelist = https://data.gov https://www.data.gov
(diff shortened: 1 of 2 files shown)


FuhuXia (approved):
🥘

FuhuXia:
Can we confirm the CORS setting does not work? I see it is working on Inventory.

image

FuhuXia:
It turns out ckan CORS config responds correctly if it is a preflight request with "Origin" header in the request. It does not expose the list of allowed origins unless your origin matches. 
https://github.com/ckan/ckan/blob/df50c365351aafbf6c929abe999d45aa08751bcb/ckan/views/__init__.py#L21-L45

Here is the sample request.
```
curl -I -H "Origin: https://gsa.gov" \
  -H "Access-Control-Request-Method: GET" \
  -H "Access-Control-Request-Headers: X-Requested-With" \
  -X OPTIONS https://inventory.data.gov
```

PR #388, opened 2021-12-20
Update rdflib to support new setuptools

nickumia-reisys:
This was triggered by `python-buildpak` supporting minimum `setuptools 59.8.0` which requires minimum `python3.7`.
```
Python Version   |  Setuptools Version  |  rdflib version  |  Working?
3.6.9            |  59.6.0              |  4.2.1           |  yes
3.7.0            |  60.0.0              |  4.2.1           |  no
3.7.0            |  60.0.0              |  4.2.2           |  yes
3.7.0            |  60.0.0              |  6.0.0           |  yes
```
- `setuptools v59.6.0`  is the last supported version for `python 3.6`
- `rdflib v6.0.0` is the latest supported major version.
- `ckanext-dcat` stills says that it relies on `rdflib v4.2.1`
- I tried to run the tests for `ckanext-dcat` with `python3.8` (that's what I have locally) and `rdflib6.0.0` and the tests fail with syntax issues with pylons...

Summary... we have to update the version of rdflib  for setuptools to be able to install it.

PR #532, opened 2022-08-24
pinning setuptools to working version per github issue: https://githu…

btylerburton:
…b.com/cloudfoundry/python-buildpack/issues/574

### DIFF
--- a/ckan/requirements.in
+++ b/ckan/requirements.in
@@ -81,7 +81,7 @@ passlib==1.7.3
 pastedeploy==2.0.1        # manually kept - remove when #4802 is complete
 pathtools==0.1.2          # via watchdog
 polib==1.0.7
-psycopg2-binary           # cloud.gov build problem
+psycopg2==2.8.6
 python-magic==0.4.15
 pysolr==3.6.0
 python-dateutil>=1.5.0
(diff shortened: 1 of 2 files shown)


btylerburton:
This passed in dev. https://github.com/GSA/catalog.data.gov/actions/runs/2921671027

btylerburton:
Also, not sure if we have a ticket for this or if I should make on for posterity. Feedback appreciated.

nickumia-reisys:
This might not be needed anymore if the problem was fixed [here](https://github.com/cloudfoundry/python-buildpack/issues/574#issuecomment-1226605056)?  The latest python buildpak should have a version of setuptools that will work out of the box now.

nickumia-reisys:
This ("sort of") happened before (a few times). Here's another example: https://github.com/GSA/catalog.data.gov/pull/388

nickumia-reisys:
Also, we don't have a ticket and it might be good to have one.  I'll just put this here (even though it's tangentially related)... it seems like our Issues and PRs are the basis for our documentation which could work (but is definitely unorthodox).  If you're suggesting creating a ticket as a documentation step that would follow our history; but should the primary documentation for this be in an issue?  I don't think so.

btylerburton:
I'm going to revert the psycopg-binary changes and run a build in dev to see if it passes.

I'd like to hear more about your thoughts on how to document this stuff. I've seen it done as a captain's log in a wiki, but that wasn't searchable enough and only really worked for smaller periods of activity.

PR #539, opened 2022-09-09
circleci is stopped

FuhuXia:
FCS resources removed, no need to keep circleci config for the sake of `FCS` branch.


nickumia-reisys (approved):
🔥

PR #540, opened 2022-09-12
update requirements for catalog to pull in changes outside pypi

btylerburton:
relates to:
- GSA/data.gov#3929

### DIFF
--- a/ckan/requirements.txt
+++ b/ckan/requirements.txt
@@ -65,7 +65,7 @@ polib==1.0.7
 psycopg2==2.8.6
 pycparser==2.21
 PyJWT==2.4.0
-pyOpenSSL==20.0.1
+pyOpenSSL==22.0.0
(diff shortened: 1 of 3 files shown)


btylerburton:
this might fail since it pulls out setuptools hard pin. let's see.

ckan/requirements.txt
```diff
@@ -83,24 +86,25 @@ requests==2.25.0
 rfc3987==1.3.8
 Routes==1.13
 rq==1.0
-# pin setuptools until we get python buildpack 1.7.58
```
robert-bryson, line 86: do we still need these? if we move the pin/comment to the requirements.in it shouldn't get blown away.

robert-bryson: nevermind, its already there.

jbrown-xentity (approved):

PR #541, opened 2022-09-12
update deps (mostly for ckanext-geodatagov updates)

robert-bryson:
related to https://github.com/GSA/data.gov/issues/3648


robert-bryson:
Basically the same as https://github.com/GSA/catalog.data.gov/pull/540

PR #542, opened 2022-09-14
Docs - System list

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/pull/3951


nickumia-reisys:
I'm against putting this in each repo because the purpose of it is for O&M which needs to be a consolidated list of systems that we track and manage.  If we scatter it between repos, it's harder to get to..

PR #545, opened 2022-09-15
Feature/auto restart fetch

jbrown-xentity:
Run fetch with an auto restart whenever there is a hard failure, so that we don't have to wait for cloud.gov to recognize the failure and restart the machine (takes at least a few minutes, sometimes takes more than 10).

Related to https://github.com/GSA/data.gov/issues/3550


nickumia-reisys:
Don't you want to borrow part of this PR?
- https://github.com/GSA/catalog.data.gov/pull/526

jbrown-xentity:
I think that's a separate update/fix, and can be done in relation to #3943 and #3944

btylerburton (approved):

PR #547, opened 2022-09-16
Adding basic auth to catalog-web

btylerburton:
Relates to:
  - GSA/data.gov#3910


proxy/.profile
```diff
@@ -0,0 +1,30 @@
...
+export SPACE_NAME=$(echo $VCAP_APPLICATION | jq -r '.space_name')
+
+echo "Setting up proxy in $APP_NAME on $SPACE_NAME"
+
+# only add auth to staging environment / otherwise turn it off
+if [[ $APP_NAME = "catalog-proxy" ]] && \
+[[ $SPACE_NAME = "staging" ]]
+then
+    echo "Setting proxy basic auth..."
+    echo "$(vcap_get_service secrets .credentials.PROXY_AUTH.USERNAME):$(vcap_get_service secrets .credentials.PROXY_AUTH.PASSWORD)" > ${HOME}/etc/nginx/.htpasswd
```
nickumia-reisys, line 24: What happens if there is a parse error here?

btylerburton: Good question. Are there any examples of parsing with a fallback in the repo?

btylerburton:
@jbrown-xentity what would be the desired action if we hit an error?

btylerburton:
Gave this some thought over the weekend and I think if the values are missing we don't want to skip adding auth, or add auth in a broken state, so might be best to fail the build. 

Will chat with @FuhuXia and @Jin-Sun-tts about this and get their input, then make changes accordingly.

FuhuXia:
Can we see it in action on `development` space?

FuhuXia (approved):

PR #548, opened 2022-09-20
restart fetcfh only when it is idling

FuhuXia:
For https://github.com/GSA/data.gov/issues/3962

Check catalog-fetch instances before restart. When none of them is generating new logs, and none of them shows CPU usage >=1%, we consider catalog-fetch is idling, can it can be restarted.

### DIFF
--- a/.github/workflows/restart.yml
+++ b/.github/workflows/restart.yml
@@ -84,7 +84,7 @@ jobs:
       - name: restart fetch
         uses: cloud-gov/cg-cli-tools@main
         with:
-          command: cf restart catalog-fetch
+          command: tools/restart_fetch.sh
           cf_org: gsa-datagov
           cf_space: prod
           cf_username: ${{secrets.CF_SERVICE_USER}}
--- a/tools/restart_fetch.sh
+++ b/tools/restart_fetch.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# Custom script to ensure fetch process is safe to restart
+# Based on https://github.com/GSA/data.gov/issues/3962
+
+###################
+# Utility Functions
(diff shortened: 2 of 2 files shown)


nickumia-reisys (changes requested):
Why make a different file?  Can't we just make a file `restart_harvester.sh` that can be passed with either `catalog-gather` or `catalog-fetch`?

nickumia-reisys (dismissed):
🕹️

nickumia-reisys:
### Added a fix for the following case

image


image

jbrown-xentity (approved):

tools/restart_harvester.sh
```diff
@@ -0,0 +1,47 @@
...
+log_last_time () {
+  date --utc --date="$(cf logs --recent $app_to_restart | tail -n 1 | awk '{split($0,time," "); print time[1]}')" +%s
+}
+
+# Restart if CPU usage is < 1
+cpu_restart () {
+  instances=$(cf app $app_to_restart | grep '^instances:' | sed 's/.*\///')
+  i=0
+  cf app $app_to_restart | tail -n $instances | awk '{ split($4,cpu,"."); print cpu[1]}' | while read -r cpu ; do
+    if [[ $cpu < 1 ]]; then
```
btylerburton, line 27: where do we get `$cpu` from?

nickumia-reisys: The `while read -r cpu` iterates through the previous command and makes the iteration available through the `cpu` variable.

btylerburton: Thanks! I feel like this line might benefit from a comment explaining the pipes.

PR #550, opened 2022-09-23
Fix restarts.... again..... restart-app-instance works

nickumia-reisys:
Normal rolling restart does not work, but restarting individual instances does work :/

### DIFF
--- a/.github/workflows/restart.yml
+++ b/.github/workflows/restart.yml
@@ -60,7 +60,7 @@ jobs:
       - name: restart catalog-web
         uses: cloud-gov/cg-cli-tools@main
         with:
-          command: cf restart catalog-web --strategy rolling
+          command: tools/restart.sh catalog-web
           cf_org: gsa-datagov
           cf_space: prod
           cf_username: ${{secrets.CF_SERVICE_USER}}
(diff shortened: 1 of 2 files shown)


Jin-Sun-tts (approved):

nickumia-reisys:
So... what happens is that the restart begins and then all of a sudden all of the apps "go down" and then it brings them back up.   After which time the apps become healthy..  It would seem like the restarts are working, there's just this weird behavior so `cf` thinks it failed, but it was just ...delayed?

### Original state
image


### Restart
image
image

### Final state
image


It seems like there's always one app that doesn't get the restart.  In this case, it seems like it is `Instance #0`.  This PR would get the action to pass, but it does not solve the underlying problem... @jbrown-xentity @FuhuXia

FuhuXia:
This instance-level restart is necessary for apps with many instances. Our catalog-web have 5. When catalog-web does a rolling restart, it restarts each instance sequentially, which make the whole deployment take so long that Cloud Foundry thinks the deployment times out.

For apps with fewer instances (1-3ish),  we can do rolling restart as usual. Especially for 1 instance, we have to do rolling restart, not restart on instance level, to prevent any down time.

PR #559, opened 2022-10-04
DB Solr Sync - Censor Logs + Add Spaces + GH Coolness!

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/3985
- #558 
- #557 

Changes:
- `monitor_cf_logs` needs another argument to only print out logs for a specified extension (ensure sensitive details are hiddne)
- Refactor `production` environment to `prod`
- Schedule DB Solr Sync for all environments

Reference:
- https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

### DIFF
--- a/.github/workflows/db-solr-sync-automated.yml
+++ b/.github/workflows/db-solr-sync-automated.yml
@@ -1,0 +1,33 @@
+name: DB Solr Sync
+
+on:
+  schedule:
+    - cron: '0 4 * * *' # every day at 4am UTC
+
+jobs:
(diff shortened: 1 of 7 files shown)


FuhuXia (approved):

jbrown-xentity (approved):
I really wish we could setup things appropriately such that we didn't have to repeat code, but github actions makes it difficult. This checks out as the best way to do it now.

FuhuXia (approved):

PR #577, opened 2022-10-10
Lint yml test, and lint fixes

jbrown-xentity:
Lint github action yml files for validity, best practices.
Uses https://github.com/adrienverge/yamllint
Documentation: https://yamllint.readthedocs.io/en/stable/index.html
I would encourage real review and discussion/clean up on this...

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -104,7 +104,7 @@ jobs:
         uses: cloud-gov/cg-cli-tools@main
         with:
           command: >
-            cf push ${{ matrix.app }} --vars-file \
+            cf push ${{ matrix.app }} --vars-file
             vars.development.yml --strategy rolling
           cf_org: gsa-datagov
           cf_space: development
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -12,9 +12,7 @@ jobs:
       - name: checkout
         uses: actions/checkout@v2
       - name: Lint GitHub Actions
-        run: |
-          cd .github/workflows
-          yamllint .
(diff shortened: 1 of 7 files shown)


.github/workflows/commit.yml
```diff
@@ -93,7 +103,9 @@ jobs:
       - name: deploy ${{ matrix.app }}
         uses: cloud-gov/cg-cli-tools@main
         with:
-          command: cf push ${{ matrix.app }} --vars-file vars.development.yml --strategy rolling
+          command: >
+            cf push ${{ matrix.app }} --vars-file \
+            vars.development.yml --strategy rolling
```
robert-bryson, line 108: It's very cool that it does this for us.

jbrown-xentity: Actually good catch, I didn't need the `\` there, command works differently. See previously working version [here](https://github.com/GSA/catalog.data.gov/blob/main/.github/workflows/db-solr-sync.yml#L40-L43)

robert-bryson: I believe the behavior of `>` in yaml can let us break some of the other long commands into their own lines too, like in those couple that you had to disable the length linting. `>` should just append each following line without adding a newline char, right?

jbrown-xentity: Possibly. Didn't have time to look into the details, just using it for what I know would work (in the cloud.gov start command field)...

robert-bryson (dismissed):

robert-bryson (approved):

jbrown-xentity:
Looks much better to me, thanks for the assist @robert-bryson!

PR #578, opened 2022-10-12
Feature/s3 creds

robert-bryson:

These darn things keep getting blown away. I just want to rebase, not sure why there are so many old commits?


proxy/nginx.conf
```diff
@@ -29,6 +29,10 @@ http {
     listen {{port}};
     root public;
     index index.html index.htm Default.htm;
+
+    location = /robots.txt {
+        alias /public/robots.txt;
```
robert-bryson, line 34: @btylerburton, I feel like you've been in the `nginx.conf` files more recently than I have, does this seem like the right place to drop this?

robert-bryson: It is within (what I think is?) the first hit server block.

`nginx-common.conf` appears to be hit after `auth` stuff is, I think. I don't think we want `robots.txt` or the sitemaps to be behind the auth, right? I might not be understanding how auth is working in this case, though.

btylerburton: check out this block then: https://github.com/GSA/catalog.data.gov/blob/main/proxy/nginx-common.conf#L19

we wanted health check not to be backed by auth and this is working

robert-bryson:
I might need some additional testing on this (the routing and the schedule action) once merged, but some of the stuff needs to be in to do it.

proxy/nginx.conf
robert-bryson: I dropped it down by the `500` block. Thanks for this!

nickumia-reisys (approved):
⛓️

PR #584, opened 2022-10-17
Limit length of archiver download

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/3986
- https://github.com/GSA/data.gov/issues/3965

References:
- https://github.com/ckan/ckanext-archiver/blob/c96e3c81bfc430cdb0372f3307c7abd4109a80f1/ckanext/archiver/default_settings.py

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ update-tracking-info:
 
 harvest:
 	# Pass any of the following arguments to run them
-	# make harvest 'fetch-consumer'
-	# make harvest 'gather-consumer'
-	# make harvest 'run'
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
It looks like we already don't use and/or leave out the "download the archived version" feature of this extension, so I think this is reasonable. Looking at the upstream code, I think this will do what we want (but it's possible that the broken link report still will not work, not fully sure).
For the record, I don't think we can or should archive the actual data/resources themselves; but it is valuable to track if the resources are available, what type they are, etc.

PR #590, opened 2022-10-18
Change logging level DEBUG to INFO

jbrown-xentity:
Related to https://github.com/GSA/data.gov/issues/3833

There are way too many logs... I don't think we need debug level info at this time, just INFO.


nickumia-reisys (commented):
Wasn't the debug level to capture login/logout events?

jbrown-xentity:
It isn't implemented yet, and we can set it to whatever level we want in the future: https://github.com/GSA/data.gov/issues/2867

PR #592, opened 2022-10-18
Change new-relic app name

jbrown-xentity:
Related to https://github.com/GSA/data.gov/issues/3991

We currently can't distinguish in the logs between catalog-web, admin, fetch, and gather. This should accomplish that.

There will be a follow up to setup catalog-web and catalog-admin with alerts as before inside New Relic (since they will have a new name), but this will allow us to target certain applications/processes for alerts better as well.


nickumia-reisys (approved):
📛

PR #600, opened 2022-10-20
Speed up pip install

nickumia-reisys:
This is ABSOLUTELY CRAZY! But it works, well.. we'll see with the whole test run, but this should install things a lot quicker


jbrown-xentity:
We may want to investigate https://github.com/GSA/data.gov/issues/2449

PR #605, opened 2022-10-21
Fix sed with / in URL

jbrown-xentity:
Fixes proxy deployment, errors with sed and url with `/`. See [recent failures](https://github.com/GSA/catalog.data.gov/actions/workflows/publish.yml). Found in recent debugging.


btylerburton (approved):

robert-bryson:
Thanks @btylerburton and @jbrown-xentity! Nice work!

PR #606, opened 2022-10-21
Revert to basic fetch command with recent fixes

jbrown-xentity:
Catalog-fetch fails to deploy, and is in a bad state. Would rather have slow fetch jobs with insight into what's happening, than fast jobs that fail all the time and we don't know why...

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -119,7 +119,7 @@ applications:
     no-route: true
     instances: ((fetch-instances))
     disk_quota: 1.5G
-    command: newrelic-admin run-program while ! ckan harvester fetch-consumer; do sleep 1; done
+    command: newrelic-admin run-program ckan harvester fetch-consumer
     health-check-type: process
     timeout: 15
     env:


btylerburton (approved):

nickumia-reisys (commented):
I'm not sure if this is a question or just comments, but the fetch process does have a pretty smart restart logic every 15 mins (so I don't think that it'll be "slow").  Did the while loop mask the error in the while loop subshell before?

jbrown-xentity:
The while loop was masking hard error failings in fetch, so it would restart automatically/quickly (instead of cloudfoundry having to recognize the failure and spin up a new machine, a difference of <1 second to restart vs 1-10 minutes). However, with recent work on https://github.com/GSA/data.gov/issues/3532, I expect less harvest failures and this becomes less necessary. With that risk calculation, I would rather have improved auditing such that we can fix some of the rest of the bugs, and stop masking the issues.

FuhuXia:
Retried in https://github.com/GSA/catalog.data.gov/pull/1384 with correct way to wrap the while loop.

PR #614, opened 2022-10-24
Update URL for staging and dev

jbrown-xentity:
In trying to clear the cache for staging, it was discovered that these do not coincide with the GSA-OPPPROD account. If you use the ones detailed in oppprod, then you're redirected to the ones specified in the catalog config (that are changed here). My understanding is that these that are being replaced are created outside the GSA environment... These should be the correct ones.

### DIFF
--- a/vars.development.yml
+++ b/vars.development.yml
@@ -15,7 +15,7 @@ gather_memory_quota: 1G
 new_relic_monitor_mode: false
 
 # use CDN domain for route-public if available, otherwise use route-external 
-route-public: d2jqk88ququ1n9.cloudfront.net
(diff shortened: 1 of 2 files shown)


robert-bryson (approved):

FuhuXia:
The two CDNs on GSA OPPPROD account were created for this ticket https://github.com/GSA/data.gov/issues/3961. This way to can use *.data.gov domain for the URL. But since that ticket is on hold, we can remove the two CDNs.

jbrown-xentity:
Check the cloudfront distributions on ssb-development for the current list, and access to clear the cache.

PR #619, opened 2022-10-25
added file and route for reg

robert-bryson:

To register catalog.data.gov to us, google needs to have this file available. Once registered, we should be able to view/trigger sitemap crawls.


btylerburton (approved):

robert-bryson:
You're fast like lightning, @btylerburton!

PR #622, opened 2022-10-26
proxypass for sitemap

robert-bryson:

Some shellcheck appeasement, but mostly added proxy pass for `/sitemap.xml` so it can be added to the Google Search Console.

### DIFF
--- a/.profile
+++ b/.profile
@@ -48,7 +48,7 @@ function vcap_get_service () {
   name="$1"
   path="$2"
   service_name=${APP_NAME}-${name}
-  echo $VCAP_SERVICES | jq --raw-output --arg service_name "$service_name" ".[][] | select(.name == \$service_name) | $path"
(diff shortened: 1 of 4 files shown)


jbrown-xentity:
This looks good, but don't we want to then hard-code the robots.txt to have `https://catalog.data.gov/sitemap.xml`, if we're doing it dynamically in nginx?

robert-bryson:
> This looks good, but don't we want to then hard-code the robots.txt to have `https://catalog.data.gov/sitemap.xml`, if we're doing it dynamically in nginx?

Good point. Does [this](https://github.com/GSA/catalog.data.gov/pull/622/commits/a3d69d91317be22235ed3c6bc694b9602d000135) seem like what you have in mind?

jbrown-xentity (approved):

PR #623, opened 2022-10-27
rolling restart for app with one instance

FuhuXia:
For https://github.com/GSA/data.gov/issues/3927

### DIFF
--- a/.github/workflows/restart.yml
+++ b/.github/workflows/restart.yml
@@ -2,6 +2,7 @@
 name: restart application
 
 on:   # yamllint disable-line rule:truthy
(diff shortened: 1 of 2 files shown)


robert-bryson:
@FuhuXia , looks good. I made [these changes](https://github.com/GSA/catalog.data.gov/pull/623/commits/4984e2e12f4f2f53e9620daa98bb87cafd90150f) to make my shellcheck happy. I think they should be fine, but I will double check as it runs.

robert-bryson (approved):

robert-bryson (approved):

PR #628, opened 2022-10-28
Use builtin nginx envsubst

nickumia-reisys:
Related to
- #626 

Reference: https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration


nickumia-reisys:
Not as simple as I thought.

PR #634, opened 2022-10-31
fixes to sitemap url

robert-bryson:

- move sitemap url constuction to nginx
- shouldn't .profile be executable?


nickumia-reisys (approved):
I don't agree with the permission change, but it's not terrible 🔎

PR #645, opened 2022-11-03
Install jq

jbrown-xentity:
image for restarts needs jq for parsing cf api calls. See https://github.com/GSA/catalog.data.gov/actions/runs/3389382767/jobs/5632421864#step:4:29

See https://github.com/GSA/catalog.data.gov/actions/runs/3389569540/jobs/5632801431#step:4:32 for working version.

### DIFF
--- a/tools/restart.sh
+++ b/tools/restart.sh
@@ -5,6 +5,9 @@
 # Input any app name
 app_to_restart=$1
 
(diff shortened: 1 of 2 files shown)


nickumia-reisys (approved):
🍰

PR #657, opened 2022-11-14
Pipenv

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/3415

References:
- https://github.com/pypa/pipenv/issues/4430
- https://github.com/pyproj4/pyproj/issues/1111
- https://github.com/pyproj4/pyproj/issues/136
- https://stackoverflow.com/a/25723191
- https://pipenv-fork.readthedocs.io/en/latest/advanced.html
- https://realpython.com/pipenv-guide/#the-pipfile
- https://github.com/pypa/pipenv/issues/3493
- https://stackoverflow.com/questions/55592586/how-to-install-update-package-with-pipenv-without-updating-the-rest-of-package
- https://docs.cloudfoundry.org/buildpacks/python/index.html#pipenv
- https://www.pivotaltracker.com/n/projects/1042066/stories/141616409
- https://github.com/cloudfoundry/python-buildpack/commit/ff9475ca9f1f752e6ba6b558f9dd6fa5056cc130
- https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/
- https://pypi.org/project/pipenv/#history
- https://stackoverflow.com/questions/57646310/is-pythons-pipenv-slow


nickumia-reisys:
Unanswered questions:
- [ ] Do we want to consolidate pipenv for cloud.gov use too?
- [ ] Why does this seem not much better than before?

ckan/requirements.txt
```diff
@@ -1,116 +1,114 @@
 alembic==1.0.0
-async-timeout==4.0.2
-Babel==2.9.1
-Beaker==1.11.0
+argparse==1.4.0
```
jbrown-xentity, line 2: Ideally this whole file is removed...

jbrown-xentity (commented):
I think we'd need to do something different here...

ckan/Dockerfile
```diff
@@ -31,11 +24,18 @@ ENV PIP_SRC=${SRC_DIR}
 ENV CKAN_STORAGE_PATH=/var/lib/ckan
 RUN mkdir -p $CKAN_STORAGE_PATH $APP_DIR
 
+ADD Pipfile Pipfile.lock ${APP_DIR}/
+ENV PIPENV_PIPFILE=/${APP_DIR}/Pipfile
+RUN pip3 install --upgrade pipenv && pipenv install --ignore-pipfile
```
jbrown-xentity, line 29: Why are we ignoring pipfile here? I think we want to use that...

ckan/freeze-requirements.sh
```diff
@@ -3,17 +3,10 @@
 set -o errexit
 set -o pipefail
 
-venv=$(mktemp -d)
+pip3 install pipenv
```
jbrown-xentity, line 6: I don't think we want this file anymore, but we should use the standard pipenv pinning and upgrading process

ckan/setup/ckan_setup.sh
```diff
@@ -29,11 +29,11 @@ do
 done
 
 echo "Enabling debug mode"
```
jbrown-xentity, line 31: See lines 19-26 in this file, I think we'll have to do something a little different...

ckan/setup/server_start.sh
```diff
@@ -10,7 +10,7 @@ DIR="$(dirname "${BASH_SOURCE[0]}")"
 # Threads RAM requirement unknown at this time
 # exec newrelic-admin run-program gunicorn -c "$DIR/gunicorn.conf.py" --worker-class gevent --paste $CKAN_INI "$@"
 if [[ "$CKAN_SITE_URL" = "http://ckan:5000" ]]; then
-  exec newrelic-admin run-program gunicorn "wsgi:application" --config "$DIR/gunicorn.conf.py" -b "0.0.0.0:$PORT" --chdir $DIR  --timeout 120 --workers 2
+  exec pipenv run newrelic-admin run-program gunicorn "wsgi:application" --config "$DIR/gunicorn.conf.py" -b "0.0.0.0:$PORT" --chdir $DIR  --timeout 120 --workers 2
 else
   exec newrelic-admin run-program gunicorn "wsgi:application" --config "$DIR/gunicorn.conf.py" -b "0.0.0.0:$PORT" --chdir $DIR  --timeout 120 --worker-class gevent --workers 4 --threads 4 --forwarded-allow-ips='*'
```
jbrown-xentity, line 15: Equivalent pipenv run needed here possibly; or maybe not? would require cloud.gov testing.

ckan/Dockerfile
nickumia-reisys: We're ignoring `pipfile` in favor of `pipfile.lock` which is the hard set of requirements.

ckan/setup/server_start.sh
nickumia-reisys: Okay, I'll look into this.

nickumia-reisys:
If this is ever re-visited for future work, there is still a lot of work to do with ensuring all of the dependencies are cloud.gov-compatible. There was no concrete path forward from the current state.

PR #660, opened 2022-11-16
Setup proxy on startup

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/4055
- https://github.com/GSA/data.gov/pull/4064


jbrown-xentity (approved):
I think this means we're giving up on the entire space being locked down, and just keeping that at the app level. That being said, I think this is the right approach.

jbrown-xentity:
> I think this means we're giving up on the entire space being locked down, and just keeping that at the app level. That being said, I think this is the right approach.

I was wrong, you can have separate egress per space by type, so we will have our custom egress over the whole `running` apps in a given space while the public egress will be over the `staging` process...

PR #664, opened 2022-11-23
Set http_proxy too!

nickumia-reisys:
Related to 
- Conversation with @FuhuXia 

I don't know if this will work because I think the `egress-proxy` only supports https connections, but it might be able to proxy non-tls traffic too?  Does someone else know? @robert-bryson @btylerburton @jbrown-xentity @FuhuXia @Jin-Sun-tts


nickumia-reisys:
This won't work @FuhuXia 

InkedInkedScreenshot 2022-11-23 132820

PR #669, opened 2022-11-29
Revert "Pin pysaml2 to working version"

nickumia-reisys:
Reverts 
- GSA/catalog.data.gov#553

Related to
- https://github.com/GSA/data.gov/issues/3980

This is testing since we are no longer pinned to Python 3.7, this might work.


nickumia-reisys:
Login still broken with the latest version.

PR #670, opened 2022-11-30
added robots.txt on development

Jin-Sun-tts:
Related to
- https://github.com/GSA/data.gov/issues/4083

catalog dev:
https://d2jqk88ququ1n9.cloudfront.net/robots.txt

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -103,6 +103,8 @@ jobs:
         include:
           - app: catalog-web
             smoketest: true
+          - app: catalog-proxy
+            robots-dev: true
     runs-on: ubuntu-latest
     needs:
       - vendor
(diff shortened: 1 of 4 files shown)


jbrown-xentity (approved):
I never would have thought to do this in this way, but it looks great!
PS you probably didn't even need the robots-dev: true field, as this workflow only ever deploys to dev.
@FuhuXia @robert-bryson @nickumia-reisys @btylerburton please note this means no more manually pushing code from your local up to development without using the github action, unless you want to manually do the copy command...

nickumia-reisys:
This might have broken login... The extra commit from me shouldn't be at the root of this PR.

nickumia-reisys:
@jbrown-xentity Out of curiosity, what would have been your alternative design for this?  Something in code vs. configuration?

jbrown-xentity:
I was picturing some kind of complex if statement in the nginx code that would return different files depending on the environment. Not positive it would have worked...

nickumia-reisys:
Oh, I see that.  Yeah, but that would have definitely been more complex.  Thanks for sharing :)

robert-bryson:
FWIW, I was imagining injecting the correct values in the `.profile` or something like that, as we had done initially with the sitemap url.

nickumia-reisys:
Just for the sake of knowledge sharing and peer-reviewing, @robert-bryson your solution would have been code changes, right?  Like parsing the environment variable into geodatagov and then running different code there?

PR #675, opened 2022-12-07
Scan for Vulnerabilities + make PRs automatically

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/3416

References:
- https://stackoverflow.com/a/5410784
- https://stackoverflow.com/a/74177583
- https://www.geeksforgeeks.org/cut-command-linux-examples/
- https://github.com/GSA/data.gov/blob/main/ckan/org_dataset_count.sh


btylerburton (approved):

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,66 @@
...
+            sed -i "/${package}.*/d" ckan/requirements.in
+            sed -i "/${package}.*/d" ckan/requirements.txt
+            echo $fix >> ckan/requirements.in
+          done <<< $(jq -r '.vulnerabilities[] | .moduleName,.fixedIn[]'
+            scan.json | sed 'N;s/\n/>=/');
+
+          make update-dependencies;
+          exit 1
+      - name: Create Pull Request
+        if: ${{ failure() }}
```
jbrown-xentity, line 39: Shouldn't we want to do this on success, not failure?

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,66 @@
...
+          do
+            package=$(echo $fix | cut -d "=" -f 1)
+            sed -i "/${package}.*/d" ckan/requirements.in
+            sed -i "/${package}.*/d" ckan/requirements.txt
+            echo $fix >> ckan/requirements.in
+          done <<< $(jq -r '.vulnerabilities[] | .moduleName,.fixedIn[]'
+            scan.json | sed 'N;s/\n/>=/');
+
+          make update-dependencies;
+          exit 1
```
jbrown-xentity, line 37: Why are we exiting `1` here?

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,66 @@
...
+
+          # Exit if no vulnerabilities
+          [[ "$(jq '.ok' scan.json)" == "true" ]] && exit 0;
+          while read -r fix
+          do
+            package=$(echo $fix | cut -d "=" -f 1)
+            sed -i "/${package}.*/d" ckan/requirements.in
+            sed -i "/${package}.*/d" ckan/requirements.txt
+            echo $fix >> ckan/requirements.in
+          done <<< $(jq -r '.vulnerabilities[] | .moduleName,.fixedIn[]'
```
jbrown-xentity, line 33: What if there's no fix, does this for loop run?

.github/workflows/snyk.yml
nickumia-reisys: The previous command succeeds if there are no vulnerabilities and fails if there are vulnerabilities.

.github/workflows/snyk.yml
nickumia-reisys: Exiting 1 to fail and ensure that the PR is created.  Otherwise, everything would return 0 and no PR would be created.

.github/workflows/snyk.yml
nickumia-reisys: If there's no fix, we are testing what will happen, most likely a bad PR will be created.. But in that case, we'd have to make a PR to ignore it,  we can add the functionality in the future.  But I feel like that doesn't happen often.

PR #709, opened 2022-12-21
perhaps i missed a slash? ngnix is weird

robert-bryson:
Related to https://github.com/GSA/data.gov/issues/4061


robert-bryson:
I don't know if this is legible, but [ngnix playground](https://nginx-playground.wizardzines.com/) says this should be working? 
image

I think with the extra `/`, nginx does not [pass along the rest of the url](https://serverfault.com/questions/379675/nginx-reverse-proxy-url-rewrite). I'm not sure, ngnix is weird.

jbrown-xentity (approved):
Wish I had more confirmation to add, but I'm not sure either. Looks good.

PR #715, opened 2023-01-06
use postgresql 12

FuhuXia:
use postgresql 12 for match the real environment


robert-bryson (approved):

robert-bryson:
@FuhuXia Out of curiosity, what is the context on this?

FuhuXia:
We are using postgresql 10 on local development but 12 in cloud.gov. I noticed some psql error when I restored some db dump from cloud.gov to local.  So why not bring local version 10 to 12 to be synchronized with the production environment.

PR #718, opened 2023-01-10
remove future and see who is complaining

FuhuXia:
For https://github.com/GSA/data.gov/issues/4145

This is trial-and-error, by removing future and adding py2 condition in the import statement where ever it complains missing future, hopefully we can becoming future-less.


btylerburton (approved):
I'm rerunning Build & Test to see if it passes. If so, let's pull off the bandaid and see what breaks.

FuhuXia:
Found `ckanext-archiver` is complaining. Made an effort to remove its dependency on `future` but then `future` is patched now. New PR is ready to do the patch. Close this one.

PR #724, opened 2023-01-11
Fix check-stuck-jobs workflow_dispatch options

Jin-Sun-tts:
removed extra options in the workflow_dispatch for check-stuck-jobs workfow.


nickumia-reisys (approved):
📇

PR #727, opened 2023-01-11
added the schedule for stuck-jobs

Jin-Sun-tts:
added the schedule for the automated check-stuck-jobs

### DIFF
--- a/.github/workflows/check-stuck-harvest-jobs-automated.yml
+++ b/.github/workflows/check-stuck-harvest-jobs-automated.yml
@@ -6,7 +6,7 @@ on:   # yamllint disable-line rule:truthy
     - cron: '0 5 * * *'  # every day at 5am UTC
 
 jobs:
-  db-solr-sync:
+  check-stuck-jobs:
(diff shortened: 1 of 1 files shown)


robert-bryson:
Considering this for a sec: why do we have [an automated](https://github.com/GSA/catalog.data.gov/blob/add-scheduled-workflow-stuck-jobs/.github/workflows/check-stuck-harvest-jobs-automated.yml) and [non-automated](https://github.com/GSA/catalog.data.gov/blob/add-scheduled-workflow-stuck-jobs/.github/workflows/check-stuck-harvest-jobs.yml) action? Wouldn't we ideally want to have one action that is responsible for both cases?

nickumia-reisys:
@robert-bryson I agree that it would be ideal if they were merged together.  The reason they have not been and cannot be is because of the parameterization of inputs.  What I would suggest in it's stead is a single workflow dispatch with all of the different ckan commands we want and then an automated one with all of the automated commands.

I was going to work on it eventually, but did not want to hold up the deployment of this feature.  We do need a more sustainable process for ckan tasks.

Issue #728, opened 2023-01-12
Stuck Harvest Jobs


nickumia-reisys:
Workflow with Issue: Check Stuck Harvest Jobs on Schedule
Job Failed: check-stuck-jobs
Last Commit: 0159561973248f6ffe71aceada7428c253e677b0
Number of times run: 1
Last run by: nickumia-reisys
Github Action Run: https://github.com/GSA/catalog.data.gov/actions/runs/3899337344

Jin-Sun-tts:
test on development with the scheduled workflow to check-stuck-jobs.

PR #729, opened 2023-01-12
Don't fail fast on stuck harvest jobs

nickumia-reisys:
Related to
- #728 
- #727


jbrown-xentity (approved):
This will make sure it runs on dev, staging, and prod (even if it fails on one of them): https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast.

PR #754, opened 2023-01-27
update requirement for ckanext-geodatagov

FuhuXia:
For https://github.com/GSA/data.gov/issues/4173

- brought changes from ckanext-geodatagov
- tested harvest results

### DIFF
--- a/e2e/cypress/support/command.js
+++ b/e2e/cypress/support/command.js
@@ -18,6 +18,21 @@ Cypress.Commands.add('check_harvest_done', (retries) => {
         });
 });
 
+Cypress.Commands.add('check_dataset_harvested', (retries) => {
+    // harvester is considered functioning if we get at least 1 dataset in. 
+    cy.reload(true)
+
+    cy.contains('#content section.module-content dl dt', 'Datasets').next().invoke('text').then($text => {
+        if ($text == '0') {
+            cy.wait(1000);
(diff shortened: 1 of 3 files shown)


nickumia-reisys:
It looks like since it has the file to do the transformation now, some harvest source is throwing a real error now.

robert-bryson (approved):

e2e/cypress/support/command.js
```diff
@@ -18,6 +18,21 @@ Cypress.Commands.add('check_harvest_done', (retries) => {
...
 });
 
+Cypress.Commands.add('check_dataset_harvested', (retries) => {
+    // harvester is considered functioning if we get at least 1 dataset in. 
+    cy.reload(true)
+
+    cy.contains('#content section.module-content dl dt', 'Datasets').next().invoke('text').then($text => {
+        if ($text == '0') {
+            cy.wait(1000);
+            cy.check_dataset_harvested(retries - 1);
```
btylerburton, line 28: If I'm reading this right, you need to put the `retries == 0` check first... 
Otherwise, when `$text == 0` & `retries == 0`, the second block will never get tripped.

FuhuXia: Good catch.

FuhuXia (commented):
fixed in https://github.com/GSA/catalog.data.gov/pull/755

PR #775, opened 2023-02-08
dont use ckan storage in our ephemeral apps

FuhuXia:
if `ckan_storage` is enabled, it allows user to upload logo image in organization creation page. We do not want to support this since all uploaded files will be gone after next restart. 

image

### DIFF
--- a/.profile
+++ b/.profile
@@ -87,7 +87,6 @@ export CKAN_SQLALCHEMY_URL=$(vcap_get_service db .credentials.uri)
 export CKAN___SQLALCHEMY__POOL_SIZE=250
 export CKAN___SQLALCHEMY__MAX_OVERFLOW=500
 
-export CKAN_STORAGE_PATH=${SHARED_DIR}/files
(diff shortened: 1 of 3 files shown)


nickumia-reisys:
Local storage is for more than that and I don't think we can disable it in this way..

FuhuXia:
Did not work out. replaced with other changes in datagovtheme repo. https://github.com/GSA/ckanext-datagovtheme/pull/153

jbrown-xentity:
It's not a bad idea to remove it, so you get an error instead of a false positive (I'll take your image, even though it'll disappear in 15 minutes). This also triggered something, the s3 filestore extension we use for catalog may have the ability to handle the organization logo. Their readme doesn't have much implementation on it, so not sure... https://github.com/keitaroinc/ckanext-s3filestore/blob/main/ckanext/s3filestore/uploader.py#L182-L186
We could verify if the images on inventory orgs is coming from s3, and if so we could implement for catalog (theoretically).

PR #798, opened 2023-02-17
Revert "CKAN 2.9.8"

nickumia-reisys:
Reverts 
- GSA/catalog.data.gov#793

There seems to be weird JS issues.

image

### DIFF
--- a/ckan/requirements.in
+++ b/ckan/requirements.in
@@ -1,6 +1,6 @@
 
 # CKAN requirements and extensions
-ckan==2.9.8
+ckan==2.9.7
 git+https://github.com/ckan/ckanext-dcat@master#egg=ckanext-dcat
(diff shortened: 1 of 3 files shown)


FuhuXia (approved):

jbrown-xentity (approved):

jbrown-xentity:
Weird. I just did a build from before this PR merge on main, and the site came up fine. Not sure what happened there. The only thing I can think of is if cloudfront or ngnix somehow cached a 404, and that caused the issue to clear up before the deployment actually finished?
image

FuhuXia:
site looks fine even before the revert PR is deployed. I haven;t done anything on the cf cache yet. so many questions. we need to test this on development and we need to turn cache CF back on to development.

PR #814, opened 2023-02-21
disable automated db-solr-sync

FuhuXia:
disable it for now so that it does not cause conflict with the initial manual db-solr-sync job, which may last for quite a few hours.


nickumia-reisys (approved):
🖌️

PR #822, opened 2023-02-24
Consolidate ckan commands 🌊 

nickumia-reisys:
Changes:
- All automated commands in one action
- All manual commands in another action

References:
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
- https://yamllint.readthedocs.io/en/stable/disable_with_comments.html

PR #832, opened 2023-02-27
Complete Automated Tasks Reorg

nickumia-reisys:
Related to
- #822 

Changes:
- Convert stuck jobs issue template to a consolidated auto tasks error issue template
  - Raise issue for `stuck jobs` if it hard fails
  - Raise issue for `tracking update` if it exceeds time specified
  - Raise issue for `db-solr-sync` if it exceeds time specified
- Add comments for all matrix configuration

References:
- https://www.xmodulo.com/measure-elapsed-time-bash.html
- https://github.com/nickumia-reisys/github-action-test/actions/runs/4285249138/jobs/7463301219
- https://docs.github.com/en/actions/learn-github-actions/variables


nickumia-reisys:
Note: If multiple tasks "fail" before they are triaged, there will only be one issue that could be related to stuck jobs and tracking update.  Although this is mostly unlikely, we can also use a different template for each type of issue which would help raise the issues separately.  I'll leave it up to the reviewer to decide if we should implement this now (or in the future).

FuhuXia (approved):

PR #838, opened 2023-02-28
Revert "check solr multiple index folders"

FuhuXia:
This reverts commit 551bf6655cff4a64640fd131e5064419961f6e8c.

this sends solr to a restart loop. It looped 15 time, took 30 mins to get to a good state. we need to understand more on the issue to come up with a better fix.


btylerburton (approved):
LGTM 👍🏼

PR #839, opened 2023-03-01
Split harvesting into separate workflow

nickumia-reisys:
Related to
- #822 

When everything was combined, it was only running things on the harvesting schedule and the other schedules were never being honored... I think this is because the total time for the automated jobs (when consolidated) was too long (>8 mins) and it was probably messing with the scheduling conditions..?


nickumia-reisys:
New developments... don't merge or approve yet.

Issue #864, opened 2023-03-11
Automated CKAN Job Error Condition


nickumia-reisys:
Workflow with Issue: 4 - Automated CKAN Jobs
Job Failed: ckan-auto-command
CKAN Command (in question): ckan geodatagov db-solr-sync
CKAN Command Schedule: 0 3 * * *
Cloud.gov Environment: prod

Last Commit: c99750338031d9bc1a481435550e8ce87481456b
Number of times run: 1
Last run by: nickumia-reisys
Github Action Run: https://github.com/GSA/catalog.data.gov/actions/runs/4395446559

nickumia-reisys:
The same out of memory issue as 
- #863 

We'll probably need to increase RAM requirements for prod?

nickumia-reisys:
This was caused by a translation mistake when refactoring the `ckan_auto.yml` file,
- https://github.com/GSA/catalog.data.gov/pull/865

nickumia-reisys:
Successful (db-solr-sync) Run: https://github.com/GSA/catalog.data.gov/actions/runs/4405522910/jobs/7716516445

nickumia-reisys:
Related to
-  - https://github.com/GSA/data.gov/issues/4224

PR #865, opened 2023-03-13
Prod auto tasks should have more ram 🦆 

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/4224
- https://github.com/GSA/catalog.data.gov/issues/864


jbrown-xentity (approved):
Will this cause us to run out of memory in the org? I know we've bumped up against the limit a few times...

Issue #882, opened 2023-03-20
Automated CKAN Job Error Condition - Sitemap Failure


nickumia-reisys:
Workflow with Issue: 4 - Automated CKAN Jobs
Job Failed: ckan-auto-command
CKAN Command (in question): ckan geodatagov sitemap-to-s3
CKAN Command Schedule: 0 2 * * *
Cloud.gov Environment: prod

Last Commit: 081e30756c70c043d8aef08981db9e79b31f87b1
Number of times run: 1
Last run by: nickumia-reisys
Github Action Run: https://github.com/GSA/catalog.data.gov/actions/runs/4464046827

nickumia-reisys:
I'm not sure how to triage or debug this, but it seems like something is wrong with the S3 bucket?

Failure [3/17](https://github.com/GSA/catalog.data.gov/actions/runs/4453161299/jobs/7821451996): Total records: 372163
Lase few logs:
```bash
2023-03-18T02:39:22.92+0000 [APP/TASK/ckangeodatagovsitemap-to-s3-4453161299-1/0] OUT 2023-03-18 02:39:22,925 INFO  [ckanext.geodatagov] 372001 to 372163 of 372163 records done.
2023-03-18T02:39:22.92+0000 [APP/TASK/ckangeodatagovsitemap-to-s3-4453161299-1/0] OUT 2023-03-18 02:39:22,926 INFO  [ckanext.geodatagov] done with sitemap/sitemap-373.xml.
2023-03-18T02:39:22.92+0000 [APP/TASK/ckangeodatagovsitemap-to-s3-4453161299-1/0] OUT 2023-03-18 02:39:22,926 INFO  [ckanext.geodatagov] Starting S3 uploads...
2023-03-18T02:39:22.92+0000 [APP/TASK/ckangeodatagovsitemap-to-s3-4453161299-1/0] OUT 2023-03-18 02:39:22,926 INFO  [ckanext.geodatagov] Setting S3 globals...
2023-03-18T02:39:23.50+0000 [APP/TASK/ckangeodatagovsitemap-to-s3-4453161299-1/0] OUT Exit status 1
```

Failure [3/18](https://github.com/GSA/catalog.data.gov/actions/runs/4458469328/jobs/7830280105): Total records: 372351 
Lase few logs:
```bash
[...]

PR #887, opened 2023-03-21
Bind catalog-gather to catalog-s3

nickumia-reisys:
Related to
- https://github.com/GSA/catalog.data.gov/issues/882


jbrown-xentity (approved):
Necessary, due to https://github.com/GSA/catalog.data.gov/blob/main/.github/workflows/ckan_auto.yml#L70

PR #896, opened 2023-03-27
no yamllint line-length check, no warnings

FuhuXia:


FuhuXia:
did not work

.github/workflows/publish.yml
```diff
@@ -1,7 +1,7 @@
 ---
 name: 2 - Publish & Deploy
 
-on:   # yamllint disable-line rule:truthy
+on:   # yamllint disable rule:line-length --no-warnings
```
nickumia-reisys, line 4: This was not the way to implement this...  It would have to be
```yaml
# yamllint disable rule:line-length --no-warnings
some:
  yaml:
    config: true

```

i.e. it needs to be on a separate line and above everything that you want to apply it to..

FuhuXia: ok, giving it another try

FuhuXia: With multiple level of indenting in our code, lines tend to get long quite easy. Breaking some long lines makes the code less readable, which defeats the purpose of using yamllint. I have no problem keep the line-length disabled, or at least use a big char limit, say 120, instead of 80.

robert-bryson: Very valid points. Do you think something [like this](https://github.com/GSA/catalog.data.gov/pull/897) would work better?

PR #898, opened 2023-03-27
input -> string

robert-bryson:
When making the yamllint changes earlier, I noticed: 
image

The [docs also seem to give the example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatch) of `string` and not `input` for `type`s. I don't see this [action running regularly](https://github.com/GSA/catalog.data.gov/actions), so I might be misunderstanding the context of this.


nickumia-reisys (approved):
📋

PR #914, opened 2023-04-05
Upgrade following cypress path

jbrown-xentity:
Found that our Cypress was way out of date, so out of date it couldn't be used with m1 chips. This should upgrade us to at least 11.x, which should solve that problem.

Utilized: https://docs.cypress.io/guides/references/migration-guide


btylerburton:
you might have to add this flag to keep tests from sandboxing

https://github.com/GSA/datagov-11ty/pull/136/files#diff-5bae461d15457335f3d6285d35d3df2fd618e583eb7460c52cd8f0de69627abcR13

jbrown-xentity:
> you might have to add this flag to keep tests from sandboxing
> 
> https://github.com/GSA/datagov-11ty/pull/136/files#diff-5bae461d15457335f3d6285d35d3df2fd618e583eb7460c52cd8f0de69627abcR13

Good call @btylerburton . I added it only when actually logging in (and we need to keep state), per best practices in [cypress docs](https://docs.cypress.io/guides/references/migration-guide#Command--Cypress-API-Changes)

btylerburton (approved):

PR #915, opened 2023-04-06
Create CSW harvest source within test org;

btylerburton:
Updates spec to create CSW harvest source within test harvest org


btylerburton:
Now the question is, How was this test passing in the first place?

jbrown-xentity (approved):

PR #921, opened 2023-04-13
Owslib fixes

btylerburton:
Related to:
- https://github.com/GSA/data.gov/issues/4231

Background:
This patches the security exception flagged by Snyk, but also breaks ckanext-spatial's support for CSW sources. The intention is to eventually remove support for CSW's entirely, but until we confirm an upgrade path for our single CSW harvest source we will upgrade the lib and live with a broken CSW pipeline.


nickumia-reisys (approved):
🧣

PR #924, opened 2023-04-17
Solr debug

FuhuXia:
look into content of `/var/solr/data/ckan/data/index.properties`
This should be the right file to tell what is the current index folder.


robert-bryson (approved):

nickumia-reisys (approved):
🎇

PR #935, opened 2023-04-27
fix sitemap cron time

robert-bryson:
The sitemap job cron times need to be matching to run. The current sitemap files haven't been updated since 2023-04-05. This PR changes the run time to 2am UTC (should be 10pm EST).


nickumia-reisys (approved):
👁️

robert-bryson:
I guess we (I) never made a bug ticket for this, but in any case, the [current sitemap](https://catalog.data.gov/sitemap.xml) has an update date from yesterday and here is [the action run](https://github.com/GSA/catalog.data.gov/actions/runs/4847787364/jobs/8638342190) for it.

PR #936, opened 2023-04-27
bump cflinux to 3->4

robert-bryson:
@FuhuXia , I'm seeing 
image
in the catalog staging logs. Do you think this is the problem?

It looks like [`cflinuxfs3` has been deprecated](https://blogs.sap.com/2023/02/16/deprecation-of-cloud-foundry-stack-cflinuxfs3-and-migration-to-cflinuxfs4/) in favor of `4`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM cloudfoundry/cflinuxfs3
+FROM cloudfoundry/cflinuxfs4
 # Specify Python version
 ARG PY_VERSION=python3.8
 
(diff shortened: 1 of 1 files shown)


nickumia-reisys:
The buildpacks support both versions as of now...
- https://github.com/cloudfoundry/python-buildpack/releases/tag/v1.8.9
- https://github.com/cloudfoundry/python-buildpack/releases/tag/v1.8.4

image


It is a good fix for the future.... But the error above showed that the old `v1.7.58` doesn't support the new cflinux4 platform.  Also note that changing this for the vendoring does not automatically select a specific version for the deployment.
- https://github.com/cloudfoundry/python-buildpack/releases/tag/v1.7.58
image

FuhuXia (approved):

PR #937, opened 2023-04-28
pinned to newer buildpack

FuhuXia:
cant use 1.7.58 any more. use 1.8.4, same as inventory.
eventually we need to get [this](https://github.com/GSA/data.gov/issues/4259) fixed and use latest python buildpack.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -4,7 +4,7 @@ applications:
   - name: ((app_name))-web
     buildpacks:
       - https://github.com/cloudfoundry/apt-buildpack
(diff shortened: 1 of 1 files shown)


nickumia-reisys (approved):
🥾

robert-bryson:
`1.8.9` is current. Did you two test anything >`1.8.4` or is that version just known to be working since it is on inventory?

nickumia-reisys:
All newer versions are going to fail because of the known issue.  We didn't test anything this time around, but I am confident that no newer version will work because of
- https://github.com/GSA/data.gov/issues/4259

PR #944, opened 2023-05-08
sync requirements.in for sqlparse

FuhuXia:
sqlparse was updated in requirements.txt but not in requirements.in

### DIFF
--- a/ckan/requirements.in
+++ b/ckan/requirements.in
@@ -97,7 +97,7 @@ simplejson==3.18.0 # ckan 2.9.5 requires 3.10.0 only
 # Following can be ignored: https://github.com/ckan/ckan/pull/4450
 # sqlalchemy-migrate==0.12.0
 SQLAlchemy==1.3.5
-sqlparse==0.4.2
+sqlparse==0.4.4
 tzlocal==1.3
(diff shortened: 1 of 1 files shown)


jbrown-xentity:
test importer should not be a required test, we should remove that. Should we also remove vendoring dependencies?

nickumia-reisys:
@jbrown-xentity What is `test importer` supposed to do?

jbrown-xentity:
> @jbrown-xentity What is `test importer` supposed to do?

It was a library used and created to pull over harvest configs from the old (2.3) catalog, and rebuild them in the new (2.8) catalog. We don't actively use it, so it being required is a mistake. I wouldn't necessarily turn them off completely, as we might need it again someday in the future.

nickumia-reisys:
Interesting... that seems like a pretty important historical piece of information. I think we should change the name to be more descriptive of that fact too.. "harvest_config_transfer" maybe?

Only @Jin-Sun-tts and @btylerburton can change that setting, but I agree it should not be required.  As far as the future of the script, I don't know if we'll ever need it to work for `2.3 to 2.8` again, but not sure how we would rewrite it for a future usecase..

PR #948, opened 2023-05-15
detect solr core lock exception and stop solr if found

FuhuXia:
For https://github.com/GSA/data.gov/issues/4138.

Check in logs file `/var/solr/logs/solr.log`  for exception `LockObtainFailedException`, once detected, stop solr. ECS will detect the service down and start a new task.


solr/solr_setup.sh
```diff
@@ -17,34 +17,33 @@ ls -lart /var/solr/data/ckan/data;
...
+if [ -f "$logfile" ]; then
+  echo "#### CORECHECK_MARKER" >> $logfile
+fi
+
+check_corelock() {
+  # check LockObtainFailedException since last check marker
+  awk -v RS='(^|\n)CORECHECK_MARKER"\n' 'END{printf "%s", $0}' $logfile | grep -q "LockObtainFailedException"
+  if [ $? -eq 0 ]; then
+    echo "Found LockObtainFailedException in log file. Quit solr."
+    solr stop -p 8983
```
nickumia-reisys, line 31: Will `root` be able to stop solr if it's being run from the `solr` user?

nickumia-reisys: Also, it seems like we need to add another marker if we find the exception too?  If we don't, will the process keep exiting?

image

nickumia-reisys: I think we need something like... I don't think root has the `solr` command registered...
```suggestion
    su -c "solr stop -p 8983" -m solr
```

Reference: https://github.com/GSA-TTS/datagov-brokerpak-solr/blob/main/terraform/ecs/provision/leader.tf#L62-L63

nickumia-reisys: @FuhuXia verified `root` has the privilege.

nickumia-reisys (approved):
Still not sure, but sure.... 😶‍🌫️

PR #950, opened 2023-05-16
make solr debugging easier for the future

FuhuXia:
By pointing to the `solr_setup.sh` script in `develop` branch, we can test and verify any future solr script changes on `develop` branch and in the real ECS environment in `development` space.

TODO
After merge, we need to run cf command to update solr service on development space. This change has no effect on staging/prod.

### DIFF
--- a/solr/service-config-development.json
+++ b/solr/service-config-development.json
@@ -6,7 +6,7 @@
     "solrCpu": 4096,
     "solrFollowerMem": 28672,
     "solrFollowerCpu": 4096,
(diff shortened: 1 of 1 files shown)


jbrown-xentity:
Hey @FuhuXia , would this change be temporary? Could you provide a bit more description about why we should do this? I'm assuming so you can rapidly prototype/test in staging, and force push to develop for quick pulls, but it's not clear.

FuhuXia:
PR description updated.

nickumia-reisys (approved):
🤌

PR #960, opened 2023-06-12
Remove old REI GA tracking code and add new GA4 data stream

dlennox24:


jbrown-xentity (approved):
I'm not sure how to verify/test this, @dlennox24 . Not sure how this works on staging/dev... But looks good from my end.

robert-bryson (approved):

dlennox24:
We won't be able to fully verify until it's deployed and the tracking code is receiving data in Google Analytics from the live site.

PR #968, opened 2023-06-21
Update requirements.txt for geodatagov relink speed

FuhuXia:
bring changes from https://github.com/GSA/ckanext-geodatagov/pull/260 and https://github.com/GSA/ckanext-geodatagov/pull/257.


nickumia-reisys (approved):
📎

PR #974, opened 2023-06-29
create our own auth_tkt cookie

FuhuXia:
For https://github.com/GSA/data.gov/issues/4371

If a request comes from `login.gov` and goes to uri `/user/me`, we assume it means a successful authentication, therefore we add the session cookie `auth_tkt` and give it a good value. The purpose of this cookie is to stop anonymous requests from accessing the admin site.


nickumia-reisys (commented):
🥈

PR #953, opened 2023-05-22
CKAN 2.10 ❗ 

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/4209
- https://github.com/GSA/data.gov/issues/4352
- https://github.com/GSA/ckan/pull/86
- https://github.com/GSA/ckanext-datajson/pull/144
- https://github.com/GSA/data.gov/pull/4284
- https://github.com/GSA/ckanext-datagovcatalog/pull/39
- https://github.com/GSA/ckanext-datagovtheme/pull/162
- https://github.com/GSA/ckanext-googleanalyticsbasic/pull/15
- https://github.com/GSA/ckanext-metrics_dashboard/pull/12
- https://github.com/GSA/ckanext-usmetadata/pull/70
- https://github.com/GSA/ckanext-geodatagov/pull/255
- https://github.com/GSA/ckanext-dcat_usmetadata/pull/308

PR #987, opened 2023-07-14
Post release

FuhuXia:
Post ckan-210 release:

This needs go after [CKAN 2.10](https://github.com/GSA/catalog.data.gov/pull/953) is merged into `main`.

Reverting https://github.com/GSA/catalog.data.gov/pull/982 and https://github.com/GSA/catalog.data.gov/pull/986 when we stop GubHub actions for any DB activity.


nickumia-reisys (approved):
🍽️

PR #999, opened 2023-07-25
forked ckanext-spatial v2.0.0 for str validator fix

FuhuXia:
For https://github.com/GSA/data.gov/issues/4392

Need the fix from https://github.com/GSA/ckanext-spatial/commits/v2-0-0-gsa to unblock harvesting.


nickumia-reisys (approved):
🎞️

PR #1014, opened 2023-08-03
Use HOST know whether we are from CDN

nickumia-reisys:


btylerburton (commented):
is this still WIP or done?

nickumia-reisys:
WIP (mostly just an example) for 
- https://github.com/GSA/data.gov/issues/4413

This may not be the solution we go with, but it's part of one direction we can take.

Issue #848, opened 2023-03-03
📌 DB Solr Sync Auditing Log


nickumia-reisys:
Workflow with Issue: 4 - Automated CKAN Jobs
Job being auditied: ckan-auto-command
CKAN Command (in question): ckan geodatagov db-solr-sync
CKAN Command Schedule: 0 3 * * *
Cloud.gov Environment: prod
Total Execution Time: 

Last Commit: b7644b9b5f5b0d4e354e66d0789308aaf4c8775b
Number of times run: 1
Last run by: nickumia-reisys
Github Action Run: https://github.com/GSA/catalog.data.gov/actions/runs/5341134106

nickumia-reisys:
Test Graph

```mermaid
gantt
    title Mismatched Datasets per day
    dateFormat  YYYY-MM-DD
    section 3/4/2023
    104(prod)            :a1, 2000-01-01, 104s
    23(staging)          :after a1, 23s
    section 3/5/2023
    104(prod)            :a1, 2000-01-01, 104s
    23(staging)          :after a1, 23s
```

hkdctol:
@FuhuXia will review and create issue.

nickumia-reisys:
Not sure why `db-solr-sync` errored out on 6/22, but maybe cloud.gov needed to terminate the task for some reason?  I think it was mostly finished doing what it needed to do.  It's just weird..

![image](https://github.com/GSA/catalog.data.gov/assets/85196563/efe5e405-edc8-4f3f-9003-95211fc4a33a)

FuhuXia:
```
5306 packages without harvest_object need to be mannually deleted
```
 hmmm... `mannually` -> `manually` typo.

This error means 5306 packages have discrepancies between db and solr, but this db-solr-sync script does not know how to handle them. Even you manually index them, CKAN will send bogus `harvest_object_id` to solr then you still end up with discrepancies again. 

The count (5306) will not go away until we do two steps.
 1. run command `ckan geodatagov harvest-object-relink`. This will fix packages that has a good but not current `harvest_object_id`. After this, another run of `db-solr-sync` or a manual reindex will fix the package. We want to run this command manually when catalog-fetch is idling. A ongoing harvest job does not like his relink script.
 
[...]

PR #1037, opened 2023-08-23
Add `harvest-object-relink` as option in manual run

nickumia-reisys:
Related to
- https://github.com/GSA/catalog.data.gov/issues/848#issuecomment-1690390834

262760760-f15feddf-5f8a-49b3-8a8c-c91149a8b53b


FuhuXia:
It is dangerous to make it GH action. People tends to run it without making sure catalog-fetch is idling.

Issue #990, opened 2023-07-18
Sitemap Generation Failure


nickumia-reisys:
Workflow with Issue: 4 - Automated CKAN Jobs
Job Failed: ckan-auto-command
CKAN Command (in question): ckan geodatagov sitemap-to-s3
CKAN Command Schedule: 0 2 * * *
Cloud.gov Environment: prod

Last Commit: 27d8e6e9612e3674cc305bd330ca3a6ca0b63212
Number of times run: 1
Last run by: nickumia-reisys
Github Action Run: https://github.com/GSA/catalog.data.gov/actions/runs/5594487267

## How to reproduce

1. `cf run-task catalog-gather --command "ckan geodatagov sitemap-to-s3" --name "sitemap-test" -m 3000M -k 2G`

## Expected behavior

Run Successful

## Actual behavior

Out of memory

## Sketch

TODO

jbrown-xentity:
cloud.gov blip, closing.

nickumia-reisys:
Failed to build sitemap last night, not sure how it ran out of memory... ?

![image](https://github.com/GSA/catalog.data.gov/assets/85196563/207f1319-d1cc-4ea3-a024-76c94cc1f894)

nickumia-reisys:
@robert-bryson The Sitemap failed to create last night too.  We should probably increase the amount of memory allocated for that action?  We should test the amount that it needs to run (by manually running it testing with different values).  @GSA/data-gov-team

nickumia-reisys:
The command fails at the 16th sitemap with `2.5G` allocated.  The command fails at the 16th sitemap with `3G` allocated.  It would seem that there is no real benefit to adding more memory.  I believe the code needs to be modified so that it frees up memory once a sitemap is generated?  Investigation/Testing needs to be done on the correct solution.

nickumia-reisys:
Scratch that `filprofiler` is not supported in `alpine`.. switching to `memray`

nickumia-reisys:
Summary of testing completed:
- New algorithm runs fine with `2G` on staging (10k file size)
- New algorithm fails at 16th sitemap file on prod with `2G` (10k file size)
- New algorithm succeeds on prod with `3G` (10k file size)
- New algorithm fails on prod with `2500M` (10k file size)
- New algorithm succeeds on prod with `2G` (5k file size)
[...]

PR #1045, opened 2023-08-29
re-enables sitemap after solr query fix

btylerburton:
related to:

- https://github.com/GSA/catalog.data.gov/issues/990
- https://github.com/GSA/ckanext-geodatagov/pull/268


FuhuXia (approved):

nickumia-reisys:
Two things:
- I was looking for when the sitemap action ran, but couldn't find it.  And then I realized that this got merged only 6 hours ago and it would have ran 10 hours ago (so... we have to wait until tomorrow for the scheduled action to run)
- I still think we should not have re-enabled the scheduled action until we saw the action being successful.  We COULD HAVE upgraded `geodatagov`, merged that and then manually triggered the action with the Github Action workflow dispatch on prod.... but oh well, next time I guess...

PR #1046, opened 2023-08-30
Add PUBLIC_ROUTE into nginx server_name

FuhuXia:
For https://github.com/GSA/data.gov/issues/4413

On CloudFront we will add `HOST` and `USER-AGENT` into the header.  We need to change nginx conf accordingly. This is the 3-step deployment.
1. Add PUBLIC_ROUTE as new host, but disable `nginx-cloudfront` check
2. TODO: apply new CloudFront settings
3. TODO: re-enable `nginx-cloudfront` check.

PR #1052, opened 2023-08-31
step 3 of nginx deployment

FuhuXia:
redo step 3 of https://github.com/GSA/catalog.data.gov/pull/1046

- renabling nginx-cloudfront
- disable basic auth on staging for now. we will keep staging down until we find a good way to add basic auth back to staging.


nickumia-reisys (approved):
🏁

PR #1088, opened 2023-10-02
exclude bots from tracking

FuhuXia:
For https://github.com/GSA/data.gov/issues/4452

Stop top bots from entering data into _tracking, responding with an emtpy page.


nickumia-reisys (approved):
🍼

PR #1107, opened 2023-10-13
reduce max-parallel for Egress check from 4 to 2

FuhuXia:
When check-egress jobs failed due to `memory_in_mb exceeds organization memory quota`, it tends to fail on the 5th or 6th job. Probably it has something to do with our `max-parallel: 4` setting. Let us see if making it 2 will alleviate the problem.

### DIFF
--- a/.github/workflows/egress_check.yml
+++ b/.github/workflows/egress_check.yml
@@ -13,7 +13,7 @@ jobs:
     environment: ${{ matrix.environ }}
     strategy:
       fail-fast: false
-      max-parallel: 4
+      max-parallel: 2
       matrix:
         command:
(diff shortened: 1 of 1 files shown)


robert-bryson (approved):

FuhuXia:
After deployment, each egress check task takes 11mins to execute, double the time it took before (5-6min). That is expected. So far no more memory error.

nickumia:
This has always been the tradeoff... total time in operation vs. total memory utilization.  I hesitate with long-running CI/CD actions because it means it isn't modular enough OR, in this case, there's special constraints that increase overhead.  I'm glad @FuhuXia made the decision 😉

PR #1112, opened 2023-10-24
fix dataset/new

FuhuXia:
For https://github.com/GSA/data.gov/issues/4161#issuecomment-1775540820

before it matches `/dataset/new*`, now it matches `/dataset/new`.

### DIFF
--- a/proxy/nginx-common.conf
+++ b/proxy/nginx-common.conf
@@ -82,6 +82,6 @@ location = /500.html {
 }
 
 # prevent users from accessing: '/dataset/new' route, 'package_create' and 'resource_create' API routes
-location ~ ^/(dataset\/new|api\/action\/package_create|api\/action\/resource_create)\/? {
+location ~ ^/(dataset\/new|api\/action\/package_create|api\/action\/resource_create)/?$ {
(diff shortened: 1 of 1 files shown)


Jin-Sun-tts (approved):

proxy/nginx-common.conf
```diff
@@ -82,6 +82,6 @@ location = /500.html {
 }
 
 # prevent users from accessing: '/dataset/new' route, 'package_create' and 'resource_create' API routes
-location ~ ^/(dataset\/new|api\/action\/package_create|api\/action\/resource_create)\/? {
+location ~ ^/(dataset\/new|api\/action\/package_create|api\/action\/resource_create)/?$ {
```
btylerburton, line 85: nice fix, fuhu!

nickumia: Did you really need to remove the `\`?  I like the end-of-line, easy fix 😉

FuhuXia: In nginx, escaping forward slash is opional. But since the leading `/` is not escaped, why escaping the trailing one.

PR #1119, opened 2023-10-26
corrects list of team-reviewers, we think

btylerburton:
if this doesn't fix the issue, we can revert

Screenshot 2023-10-26 at 2 10 12 PM


btylerburton:
root cause: https://github.com/peter-evans/create-pull-request/issues/1638#issuecomment-1491165160

TL;DR - exclude org from your team name

FuhuXia (approved):

PR #1139, opened 2023-11-09
switch map tile server to openstreetmap.org

FuhuXia:
For https://github.com/GSA/data.gov/issues/4493
- changed to openstreetmap cached map tiles
- make sure requests to from referer *.data.gov *admin-datagov.app.cloud.gov


FuhuXia:
verified on development space

Jin-Sun-tts (approved):

PR #1156, opened 2023-11-17
better grep

FuhuXia:
Started to see this error in the **monitor task output** step,as in https://github.com/GSA/catalog.data.gov/actions/runs/6901149024/job/18775455943
```
Running command: tools/monitor_cf_logs.sh catalog-admin $INSTANCE_NAME
grep: unrecognized option: line-buffered
BusyBox v1.36.1 (2023-11-06 11:32:24 UTC) multi-call binary.
```

It means something in the `cloud-gov/cg-cli-tools` has been changed recently that the built-in `grep` does not support `line-buffered` any more. So we install a new grep.

### DIFF
--- a/tools/monitor_cf_logs.sh
+++ b/tools/monitor_cf_logs.sh
@@ -7,6 +7,9 @@
 app_to_monitor=$1
 task_to_monitor=$2
 
+# install a better version of grep which supports --line-buffered
+apk add grep 
+
 while read -r line ; do
   echo "$line"
   if echo "$line" | grep "OUT Exit status 0"; then


jbrown-xentity (approved):

jbrown-xentity:
I'm guessing it's [this change](https://github.com/cloud-gov/cg-cli-tools/pull/20/files). @FuhuXia , can you make an upstream PR so this addition isn't necessary moving forward? I don't like the idea of relying on an installation in a run script...

FuhuXia:
@jbrown-xentity We can give the intermediate upstream a try. Let us hope `cloud-gov/cg-cli-tools` do not kick back and push me go to the top upstream `cloudfoundry/cli`. :rofl:

robert-bryson (approved):

FuhuXia:
[Request](https://github.com/cloud-gov/cg-cli-tools/pull/23#issuecomment-1817218527) sent to upstream. In the meanwhile let us merge this PR so that we can continue monitoring cf output.

Issue #1159, opened 2023-11-21
Egress Check Failed


btylerburton:
Workflow with Issue: 6 - Check Egress Operation
Job Failed: egress-check
Type of Failure: BadDomainTest || curl -I -L --http1.1 https://yahoo.com | grep \"HTTP/1.1 403 Forbidden\"
Location of Failure: catalog-web
Cloud.gov Environment: development

Last Commit: 76b096b55da97b999b0d13baa697197319b2ba95
Number of times run: 1
Last run by: rshewitt
Github Action Run: https://github.com/GSA/catalog.data.gov/actions/runs/6949282581

rshewitt:
testing. closing.

PR #1160, opened 2023-11-21
remove auto-assignee from issue.

rshewitt:
related to [#4536](https://github.com/GSA/data.gov/issues/4536)

- removed assignee from the create-issue step
- verified by manual run of action
- issue created was verified. [link](https://github.com/GSA/catalog.data.gov/issues/1159)


FuhuXia (approved):

FuhuXia:
Agree. We don't need a ticket to be assigned to anyone.

PR #1233, opened 2024-01-29
block bad bots from 8.219.*.* and 8.222.*.*

FuhuXia:
block some bad bots.

details in 
https://gsa-tts.slack.com/archives/C2N85536E/p1706288852750289


robert-bryson (approved):

FuhuXia:
deployed and verified, all bad bat bot traffic are responded with 403. The traffic are still coming from time to time but they never reach catalog-web.

image

PR #1241, opened 2024-02-02
open development and staging for package create

FuhuXia:
during development we want to open development and staging to package create


rshewitt:
awesome, thanks for doing this! I didn't see this before I pushed my change to `develop`.

rshewitt (approved):

PR #1249, opened 2024-02-09
Static pages

FuhuXia:
Part of
- https://github.com/GSA/data.gov/issues/3907

Changes:
- create template.html from a catalog 404 page
- simplify the template to trim off unnecessary assets
- add static assets (img, css, fonts)
- change 500.html to use the new look.


FuhuXia:
deployed to development
https://catalog-dev.data.gov/500.html

proxy/public/static-assets/css/d05bf0e7_fontawesome.css
```diff
@@ -0,0 +1,7003 @@
...
+  font-family: "Font Awesome 5 Free";
+  font-display: block;
+  font-weight: 900;
+  src: url("/static-assets/fonts/fa-solid-900.woff2") format("woff2"), url("/static-assets/fonts/fa-solid-900.ttf") format("truetype"); }
+
+@font-face {
+  font-family: "Font Awesome 5 Free";
+  font-display: block;
+  font-weight: 400;
+  src: url("../../base/vendor/fontawesome-free/webfonts/fa-regular-400.woff2") format("woff2"), url("../../base/vendor/fontawesome-free/webfonts/fa-regular-400.ttf") format("truetype"); }
```
btylerburton, line 6982: there's quite a few of these relative path links in the CSS and JS. How does the page look when it's rendered?

FuhuXia: these css relative links are not used. I only updated the minimal, actually being used, css links.

btylerburton: As long as the page looks good. There's a bunch of relative links in the JS file as well, but those are more than likely to unused modules.

btylerburton (approved):
⚡

proxy/public/static-assets/css/d05bf0e7_fontawesome.css
FuhuXia: Yep. The way I use to tell whether we are missing some assets is looking at the network tab in the inspector. I did not see any 404. All necessary assets are identified and saved a local copy.

image

PR #1280, opened 2024-03-25
CKAN 2.10.4

FuhuXia:
For https://github.com/GSA/data.gov/issues/4560

- use ckan core 2.10.4
- update requirement.txt
- update saml cert
- add `frange` to search setting


jbrown-xentity (approved):
dev looks good, though I would love if there was some data in there. Tests are passing... Curious what you did to get past that first blocker @FuhuXia

PR #1283, opened 2024-03-29
update dependencies, update snyk ignore list

FuhuXia:
keep things update to date

### DIFF
--- a/ckan/.snyk
+++ b/ckan/.snyk
@@ -37,34 +37,6 @@ ignore:
           https://github.com/GSA/data.gov/issues/4303
         expires: 2024-03-31T16:20:58.017Z
         created: 2023-05-08T16:20:58.023Z
-  SNYK-PYTHON-WERKZEUG-6041510:
-    - '*':
-        reason: >-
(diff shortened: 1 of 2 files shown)


ckan/.snyk
```diff
@@ -37,34 +37,6 @@ ignore:
           https://github.com/GSA/data.gov/issues/4303
         expires: 2024-03-31T16:20:58.017Z
         created: 2023-05-08T16:20:58.023Z
-  SNYK-PYTHON-WERKZEUG-6041510:
```
btylerburton, line 40: I don't see a werkzeug version bump. 

Did this resolve the other werkzeug exception on L19 as well?

What about the Flask one on L33?

btylerburton:
Any changes to requirements.in needed?

ckan/.snyk
FuhuXia: did not evaluate what is fixed. Just evaluated what is not needed. I removed this exception, snyk did not complain.

btylerburton (approved):
LGTM

PR #1304, opened 2024-04-11
use forked ckanext-harvest

FuhuXia:
For https://github.com/GSA/data.gov/issues/4505

Bringing a [change](https://github.com/GSA/ckanext-harvest/commit/c355ed638be1b66bbaa06d736a21298cc3fd0a73) from GSA ckanext-harvest fork before an known issue is fixed on upstream. While the issue needs more time to file/resolve upstream, this PR will fix catalog and eliminate the main cause for lost-harvest-object issue.

### DIFF
--- a/ckan/requirements.in
+++ b/ckan/requirements.in
@@ -1,7 +1,7 @@
 # CKAN requirements and extensions
 ckan==2.10.4
 git+https://github.com/ckan/ckanext-dcat@master#egg=ckanext-dcat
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
Discussed in person. Looks like https://github.com/ckan/ckanext-spatial/blob/master/ckanext/spatial/harvesters/base.py#L683-L726 is where this bug is starting from; the logic seems unnecessary (creating new, matching new to old, delete old, update index for new vs just updating old), however this patches it in a safe way that is testable and repeatable. To follow up on fixing bug at source in another PR/ticket.

PR #1307, opened 2024-04-15
more locales to filter out

FuhuXia:
CKAN 2.10.0 [added](https://github.com/ckan/ckan/commit/5bc0bcc692ee02b1766af39ee8a8dba6b8139896) locales `bs` and `no`.

We received tons of requests to `/no/*`, 2 millions in a month, peaked at 10 per second.

image

### DIFF
--- a/ckan/requirements.in
+++ b/ckan/requirements.in
@@ -1,7 +1,7 @@
 # CKAN requirements and extensions
 ckan==2.10.4
 git+https://github.com/ckan/ckanext-dcat@master#egg=ckanext-dcat
--e git+https://github.com/ckan/ckanext-harvest.git@v1.5.6#egg=ckanext-harvest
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
Good catch @FuhuXia

jbrown-xentity:
Bummer @FuhuXia , looks like there's an unrelated problem with building the efs tools used in the docker image... Looks like this will have to be addressed, as the SOLR docker image won't work until we get this fixed? https://github.com/GSA/catalog.data.gov/blob/main/solr/Dockerfile#L13-L18

FuhuXia:
efs-utils v2.0.0 was released 3 days ago. Need to update install steps.

jbrown-xentity (approved):
Thanks @FuhuXia

PR #1308, opened 2024-04-15
4567 - Pointing datagovtheme to datalayer pr to test GTM ingestion

dlennox24:
https://github.com/GSA/data.gov/issues/4567


FuhuXia:
No need to create PR to  change the `develop` branch. You can just overwrite it.

jbrown-xentity:
@dlennox24 this branch isn't protected, you can use something like https://stackoverflow.com/a/5740047 to push your local onto the develop branch. Best practice is to post a notice on slack for other team members to know what's going on, in case they are trying to deploy changes as well.

PR #1316, opened 2024-04-23
fix sqlparse and gunicorn version

Jin-Sun-tts:
fixes for two failed VULNERABILITY check from snyk. 
https://github.com/GSA/data.gov/issues/4705


btylerburton (dismissed):

btylerburton:
Can't see snyk check as I'm on my phone but hopefully that's easy to mitigate.

FuhuXia (approved):

PR #1376, opened 2024-06-17
Check usage ssl_free_buffers

Jin-Sun-tts:
Added an option in the Makefile to check for vulnerable function SSL_free_buffers usage.

Related issues:
https://github.com/GSA/data.gov/issues/4781
https://github.com/GSA/data.gov/issues/4782

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -55,6 +55,15 @@ jobs:
       - name: test_importer
         run: make test-import-tool
 
+  test_vulnerability:
+    name: test vulnerability
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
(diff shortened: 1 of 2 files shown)


btylerburton:
So this test is checking for the presence of this `SSL_free_buffers` call in our dependencies? Should we be running it after every `build` or `update_dependencies`?

btylerburton (approved):

Jin-Sun-tts:
@btylerburton and @FuhuXia , added this step in the github workfile commit, it takes about 40s.

FuhuXia:
40s is a little too much , but since it runs paralleled to the 10-min Build and Test, this 40s does not slow down anything.

FuhuXia (approved):

PR #1394, opened 2024-07-08
Bump datagov theme to 0.2.30

tdlowden:


btylerburton (approved):

FuhuXia:
Should we get this into catalog before pen test is started, or wait until it is finished.

tdlowden:
@FuhuXia if pentest has not yet started, definitely would like to merge this. Without it, not sure we can progress on key aspects of metrics work

PR #1417, opened 2024-07-30
Update ckan.ini

jbrown-xentity:
Add necessary config items from ckan2.10 changelog: https://docs.ckan.org/en/2.10/changelog.html#migration-notes-2-10

Related to https://github.com/GSA/data.gov/issues/2138

### DIFF
--- a/ckan/setup/ckan.ini
+++ b/ckan/setup/ckan.ini
@@ -42,6 +42,8 @@ beaker.session.secret = TShFJxS41xNdVJAxQsoIEm5zu
 beaker.session.type=ext:database
 #beaker.session.url=postgresql://ckan:ckan@db/ckan
 beaker.session.cookie_expires=true
+beaker.session.secure = True
+beaker.session.samesite = Strict
 beaker.session.url = $CKAN___BEAKER__SESSION__URL
 beaker.session.timeout=900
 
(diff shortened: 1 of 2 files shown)


btylerburton:
Do you think we need this signing key, too?

```beaker.session.validate_key = CHANGE_ME```

btylerburton:
Same failing test. I'm not sure if anyone has tried running this locally to repro the results.

rshewitt:
> Same failing test. I'm not sure if anyone has tried running this locally to repro the results.

i'll look further into this

rshewitt:
for some reason cypress isn't able to find the feedback modal where `class=fba-modal-dialog`. @btylerburton do you see an issue with the new element being selected? I'm concerned about, what looks like, a generated id appended to the end of the touchpoint class name ( i.e. #touchpoints-form-`fd986495` ) and whether this is a viable name to use

rshewitt:
omg `#` vs `.` . i'm not paying attention.

btylerburton:
I think `fd986495` is touchpoints id for our form so it probably wouldn't change. That said, we can look for the modal by class instead of id and the test will be a little less brittle.

jbrown-xentity:
@GSA/data-gov-dev-team if someone could approve this, we could get this rolled out asap... And thanks for the test fixes @btylerburton and @rshewitt !

btylerburton (approved):

PR #1418, opened 2024-07-31
Change cookie to lax

jbrown-xentity:
Possibly due to redirects


btylerburton (approved):

btylerburton:
For posterity...
Probably need to take a look at our login tests/fixtures as well...

https://github.com/GSA/catalog.data.gov/blob/main/e2e/cypress/integration/login.cy.js

PR #1449, opened 2024-09-03
ckan 2.10.5

FuhuXia:
For https://github.com/GSA/data.gov/issues/4854
Upgrade ckan core from 2.10.4 to 2.10.5.


FuhuXia:
Verified on catalog-dev.

robert-bryson (approved):

PR #1460, opened 2024-09-23
keep facebook out of tracking stats

FuhuXia:
NewRelic dashboard shows facebook bots are parsing JS and accessing /_tracking. To keep _tracking targeted for end user traffic only, we want to ignore facebook traffic. 

image


btylerburton (approved):
Wow, that's a huge slice of the pie 🥧!!

PR #1462, opened 2024-09-23
omit prod from action matrix; scale gather and fetch down to 0

rshewitt:
related to [#4880](https://github.com/GSA/data.gov/issues/4880)


FuhuXia:
Set 0 instance might have issues to get deployed as found in https://github.com/GSA/catalog.data.gov/pull/1439. But we can give it a try first.

FuhuXia (dismissed):

FuhuXia (approved):

PR #1465, opened 2024-09-25
update ckanext-datajson for max_resource_count

FuhuXia:
Bring changes from
- https://github.com/GSA/ckanext-datajson/pull/158
- https://github.com/GSA/ckanext-datajson/pull/159

Set optional config `ckanext.datajson.max_resource_count` to be 1500.

### DIFF
--- a/.profile
+++ b/.profile
@@ -141,6 +141,9 @@ export CKANEXT__DATAGOVCATALOG__ADD_PACKAGES_TRACKING_INFO=false
 # Render recent view using AJAX call to boost page loading speed
 export CKANEXT__DATAGOVTHEME__JS_RECENT_VIEW=true
 
+# Max number of resources to be allowed in a dataset to be harvested
+export CKANEXT__DATAJSON__MAX_RESOURCE_COUNT=1500
+
 # Set up the collection in Solr
 echo Setting up Solr collection
 export SOLR_COLLECTION=ckan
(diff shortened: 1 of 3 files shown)


FuhuXia:
Verified on catalog-dev. A error is generated for a record with 10k+ resources. 

image

.profile
```diff
@@ -141,6 +141,9 @@ export CKANEXT__DATAGOVCATALOG__ADD_PACKAGES_TRACKING_INFO=false
 # Render recent view using AJAX call to boost page loading speed
 export CKANEXT__DATAGOVTHEME__JS_RECENT_VIEW=true
 
+# Max number of resources to be allowed in a dataset to be harvested
+export CKANEXT__DATAJSON__MAX_RESOURCE_COUNT=1500
+
```
jbrown-xentity, line 146: Any reason we can't add this to the ckan.ini file? I think this would be the safer place; that way it will be setup by default. I don't see a reason to have a different number between local dev and cloud.gov deployment...

jbrown-xentity (approved):
This will work, one suggested improvement...
The main reason is the less we do in the `.profile`, the faster startup should be.

.profile
FuhuXia: Tried that before, moved newly created config into ckan.ini but it had no effect.
Just tried again with https://github.com/GSA/catalog.data.gov/commit/50e405594ab9c334cb8232137540f029560a3753, same result. No effect when set in ckan.ini.

FuhuXia: oops, looks like there is a typo in the last try. trying again.

PR #1510, opened 2024-12-28
Do smoketest when scaling catalog-web

nickumia:
Apparently it hasn't been causing issues, but it seems like a safe thing to do in any case 🤷‍♀️ 

image


FuhuXia:
In order to let a **LEGIT** PR from a fork to use our GitHub secret, we can replicate the PR into a **DRAFT** PR and run github action there. 

Steps:
```
git fetch origin pull/1510/head:add-smoketest
git push origin add-smoketest
```

FuhuXia (approved):

Issue #1514, opened 2025-01-04
Automated CKAN Job Error Condition


btylerburton:
Workflow with Issue: 4 - Automated CKAN Jobs
Job Failed: ckan-auto-command
CKAN Command (in question): ckan geodatagov check-stuck-jobs
CKAN Command Schedule: 30 6 * * *
Cloud.gov Environment: prod

Last Commit: 8ed9eab28b78bdecff15f63f4e1954648b87451a
Number of times run: 1
Last run by: btylerburton
Github Action Run: https://github.com/GSA/catalog.data.gov/actions/runs/12608304457

FuhuXia:
harvest source /harvest/ncddc-oer is getting stuck due to fetch process crash.

Need to take a closer look and raise an issue on it.

![image](https://github.com/user-attachments/assets/7a957f6d-b981-4fbb-80f4-fe683ad95681)

PR #1516, opened 2025-01-09
500M for catalog-fetch

FuhuXia:
Noticed OOM for catalog-fetch process.
https://github.com/GSA/catalog.data.gov/issues/1514

bumping the mem from 300M to 500M resolved the issue.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -122,6 +122,7 @@ applications:
     no-route: true
     instances: ((fetch-instances))
     disk_quota: 1.5G
+    memory: ((fetch_memory_quota))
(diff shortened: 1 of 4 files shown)


jbrown-xentity (approved):

jbrown-xentity:
Are we sure we want to downgrade the instances from 4 to 3?

FuhuXia:
> Are we sure we want to downgrade the instances from 4 to 3?

Not really sure, but we can give 3 a try and check one month later to see the impact. As of now I can tell fetch process is idling for the most of time. 

image

PR #1551, opened 2025-02-18
Only log extension messages at the INFO level

neilmb:
The change to `ckanext-geodatagov` didn't stop our loud "Added FQ to collection_package_id" because we were logging `ckanext` at the DEBUG level in `ckan.ini`. This changes that file to log messages from extensions at the `INFO` level.

I pushed this manually to staging and it appears to have stopped the desired messages from making it to New Relic:

Screenshot 2025-02-18 at 3 01 42 PM

### DIFF
--- a/ckan/setup/ckan.ini
+++ b/ckan/setup/ckan.ini
@@ -350,7 +350,7 @@ qualname = ckan.model
 propagate = 0
 
 [logger_ckanext]
-level = DEBUG
+level = INFO
 handlers = console,consoleerror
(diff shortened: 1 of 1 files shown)


jbrown-xentity (approved):
This is probably long overdue, although this will change the logging info for all ckan extensions across all catalog instances (including harvester). However we never should have had debug level logging on at all, so I believe this is good to merge.

btylerburton (approved):

PR #1592, opened 2025-04-07
Cypress tests for catalog-next

neilmb:
For issue
- https://github.com/GSA/data.gov/issues/5137
- https://github.com/GSA/data.gov/issues/5136

This ports the `e2e/cypress/integration` tests from `main` into `catalog-next`. We no longer have actual harvests driving the tests, so each test gets its own `before()` and `after()` calls to set up and tear down orgs/datasets/etc. for that test. To avoid name collisions, we liberally use a `randomSlug()` helper function in the names.

Harvest-specific tests have been dropped, including a pair of very specific parsing tests from `spatial.cy.js`. I expect that these are being tested now in the harvest app.

The Docker compose setup was changed to use Docker's new-ish "profile" system so that all of the services can go in the same `docker-compose.yml` file and test resources won't be started unless the `--profile test` option is provided.


e2e/cypress/integration/dcat.cy.js
```diff
@@ -0,0 +1,26 @@
...
+      cy.delete_dataset(packageId);
+      cy.delete_organization(orgId);
+    });
+
+    it('Datasets have an rdf endpoint', () => {
+        cy.request(`/dataset/${packageId}.rdf`).should((response) => {
+            expect(response.status).to.eq(200);
+            expect(response.body).to.contain('dcat:Dataset');
+        });
+    });
```
jbrown-xentity, line 25: Good test!
Should we test that the link to the RDF is in the HTML?

e2e/cypress/integration/spatial.cy.js
```diff
@@ -0,0 +1,69 @@
...
+                id: packageId,
+                groups: [{ name: group.name }],
+            },
+        }).should((response) => {
+            expect(response.body).to.have.property('success', true);
+            expect(response.body.result.groups[0]).to.have.property('name', group.name);
+        });
+        cy.delete_dataset(packageId);
+        cy.delete_organization(orgId);
+    });
```
jbrown-xentity, line 68: This probably shouldn't be in the spatial, probably should be in organizations or a new "group" file?

e2e/cypress/integration/spatial.cy.js
```diff
@@ -0,0 +1,69 @@
...
+        cy.get('#draw-map-container')
+            .trigger('mousedown', { which: 1 })
+            .trigger('mousemove', { clientX: 500, clientY: 153 })
+            .trigger('mouseup');
+        cy.hide_debug_toolbar();
+        // click the apply button then on the next redirected page find the box
+        // on the map and content in the body
+        cy.get('.modal-spatial-query .modal-footer').find('[class="btn btn-primary apply"]').click();
+        cy.get('#dataset-map-container').find('svg.leaflet-zoom-animated');
+        cy.contains(/No datasets found/);
```
jbrown-xentity, line 36: I would like to add a complete working test here, maybe adding a dataset with spatial data so that the spatial search returns a dataset?

jbrown-xentity (approved):
Overall great! A few minor quibbles...

e2e/cypress/integration/dcat.cy.js
neilmb: I don't see that there are any links to the RDF document in our dataset pages. Were you expecting there to be one?

e2e/cypress/integration/spatial.cy.js
btylerburton: +1 on this enhancement

btylerburton (approved):
Looks great! No real quibbles.

FuhuXia:
The tests pass on the first run. But when I run for the 2nd time by `make clean build test`, it always gives this error:
```
Error response from daemon: network 7dc0d8d2bf6b001311516b57e9cecd1d2cb42b91bdcfa1751d24bd4f7819e0c7 not found
make: *** [Makefile:47: test] Error 1
```

I have to run some other commands to nuke docker before i can run test again.

neilmb:
I'm trying to reproduce the cypress test failures here to be able to see logs of the ckan and solr services to see why they are throwing 400 errors.

PR #1601, opened 2025-04-23
Script to fix the admin user on Solr followers

neilmb:
In pairing we discovered that a sequence of curl commands could be used to fix Solr clusters with followers that are not configured correctly. This makes a shell script to do all of those commands if necessary.


neilmb:
re: GSA/data.gov#5185

jbrown-xentity (approved):
LGTM!

PR #1615, opened 2025-05-08
use latest rust

FuhuXia:
To addess issue observed in 
- https://github.com/GSA/data.gov/issues/5227#issuecomment-2864272059

EFS requires rustc 1.82.0 or newer. The one comes with ubuntu is 1.75.0.


FuhuXia:
Please approve but don't merge.
Merge will overwrite our nginx patch to block bots.

Jin-Sun-tts (approved):

neilmb:
@FuhuXia do we need to re-tag the generated Solr image that this published and update the brokerpak to use that Solr image when it allocates our ECS clusters?

PR #1627, opened 2025-05-21
update datagovtheme and change config for user session

FuhuXia:
For https://github.com/GSA/data.gov/issues/5253

- bring in session fix code from ckanext-datagovtheme
- allow 15 mins idle time


jbrown-xentity (approved):
LGTM

PR #1669, opened 2025-08-04
remove code not used; patch evel()

FuhuXia:
- Remove not-used code that is found vulnerable by snyk code analysis
- remove vulnerable eval function. 

snyk findings screenshots
image
image


rshewitt (approved):
`ast.literal_eval` seems like a more secure alternative to `eval`. chatted with @FuhuXia about removing the test import tool.

PR #1716, opened 2025-09-04
Update to auto-deploy and allow jobs to run on catalog-beta

jbrown-xentity:
Discussed with @FuhuXia on Slack.
These changes allow us to verify various jobs and processes before cutover, and keep things on auto-deployment to remove the need for a manual push.

When we cutover, we should eventually change all these catalog-next and catalog-beta references to catalog, but that should be a simple change (and we should scan the whole repo anyways, as various services will need to change).

### DIFF
--- a/.github/workflows/ckan.yml
+++ b/.github/workflows/ckan.yml
@@ -16,7 +16,7 @@ on:
         description: 'App to run on:'
         required: true
         type: string
-        default: 'catalog-admin'
+        default: 'catalog-next-admin'
       command:
         description: 'Command to run:'
         required: true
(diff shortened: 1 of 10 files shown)


FuhuXia:
- catalog-next branch is falling behind main on secririty patch. So we will need another effort to bring catalog-next security patch up-to-date by cherry-picking from main.
- running sitemap generation on catalog-next-admin may end up with weird domain name in the sitemap files.
- cron defined in non-default branch wont get run. so we need manual trigger the tasks, or run them in cf tasks, in order to achieve the goal to verify things before cutover.

FuhuXia (approved):

PR #1751, opened 2025-12-26
3 fetch instances

FuhuXia:
Restore the fetch instance count to 3. It was temporarily reduced to 1 during an unsuccessful attempt to mitigate duplicate records.


jbrown-xentity (approved):
LGTM

PR #1789, opened 2026-03-26
remove web instance scaling

FuhuXia:
The catalog-web is running with 5 instances.  it is pretty stable these days. We can disable this action and reduce some GH usage.


neilmb (approved):
Very reasonable.

PR #1795, opened 2026-04-16
catalog goes offline

FuhuXia:
Take catalog offline by scaling all app to 0 instances, and disabling all cron tasks.


cmhedrickREI (approved):
Looks great!

